PR #7, opened 2021-04-17
Allow user to disable fancy UI

phikal:
Hope this is self-explanatory. Not everyone likes fancy UIs.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -55,6 +55,10 @@
   "Enable cycling for `corfu-next' and `corfu-previous'."
   :type 'boolean)
 
+(defcustom corfu-fancy t
+  "Show the scroll bar and the border in popups."
+  :type 'boolean)
+
 (defgroup corfu-faces nil
   "Faces used by Corfu."
   :group 'corfu
@@ -178,7 +182,7 @@ If `line-spacing/=nil' or in text-mode, the background color is used instead.")
   "Show LINES as popup at POS, with IDX highlighted and scrollbar between LO and LO+BAR."
   (let* ((size (corfu--char-size))
          ;; XXX Deactivate fancy border on terminal or if line-spacing is used
-         (fancy (and (not line-spacing) (display-graphic-p)))
+         (fancy (and corfu-fancy (not line-spacing) (display-graphic-p)))
          (lborder-curr (corfu--border (car size) (cdr size) 1 'corfu-border 'corfu-current))
          (rborder-curr (corfu--border (car size) (cdr size) -1 'corfu-border 'corfu-current))
          (rbar-curr (corfu--border (car size) (cdr size) (- (ceiling (car size) 3))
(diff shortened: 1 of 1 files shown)


minad:
But you still want to see the wide scrollbar? You only dislike the thin borders/thin scrollbar indicator?

phikal:
On my system it also hid the scroll bar, as intended?

minad:
No, it should not do that. Either you get the "fancy" thin border/thin scrollbar or you get the wide scrollbar.

phikal:
Daniel Mendler ***@***.***> writes:

> No, it should not do that. Either you get the "fancy" thin border/thin
> scrollbar or you get the wide scrollbar.

Ah, it must be that a previous customization was still in effect, that
didn't render the scrollbar. I will amend the commit to add that too.

-- 
	Philip K.

corfu.el
```diff
@@ -55,6 +55,10 @@
   "Enable cycling for `corfu-next' and `corfu-previous'."
   :type 'boolean)
 
+(defcustom corfu-fancy t
```
minad, line 58: `corfu-scrollbar` as a better name?

corfu.el
```diff
@@ -220,8 +224,11 @@ If `line-spacing/=nil' or in text-mode, the background color is used instead.")
                         (if (and lo (<= lo row (+ lo bar)))
                             (if (= row idx) rbar-curr rbar)
                           (if (= row idx) rborder-curr rborder))
-                      (propertize " " 'face (if (and lo (<= lo row (+ lo bar)))
-                                                'corfu-bar 'corfu-border))))))
+                      (propertize " " 'face (cond ((and lo (<= lo row (+ lo bar)))
```
minad, line 227: There is an easier way to do this. Just set lo=nil if you want to disable the scrollbar.

minad:
Hmm, but you also want to hide the thin border probably? Maybe this option is not good as is. There are multiple possibilities:

* only thin border
* only thin indicator
* thin border+thin indicator
* thick indicator/right border

Given that the display code is a bit of a mess, this is low priority for now. Let's see first where this develops. Maybe I will drop overlays altogether and go with something like posframe/postip in the end.

corfu.el
phikal: I thought about that, but it also hides the border so I decided to stay with fancy.

phikal:
> Maybe I will drop overlays altogether and go with something like posframe/postip in the end.

Using child frames sounds like a better idea in general. It might even be possible to not have the completion pop up at point but in a side frame.

minad:
> Using child frames sounds like a better idea in general. It might even be possible to not have the completion pop up at point but in a side frame.

If you are interested in prototyping this it would be great. I haven't used child frames before. But we would probably still keep the overlay code around for tui and depending on preference of the user. However I would drop the fancy borders given an alternative child frame display function.

minad:
I just found this https://github.com/company-mode/company-mode/issues/745#issuecomment-351875967. I may give this a try.

phikal:
I haven't used child-frames for anything serious either, but I'll take a look.

minad:
There is one serious issue which I had observed once before when I tried posframe a while ago. The internal border color cannot be set. It does not update, see https://github.com/tumashu/posframe/issues/74. Given that child frames are still buggy I am less enthusiastic about this now.

~~~ elisp
      (set-face-background 'internal-border "red" my-tooltip-child-frame)
      (when (facep 'child-frame-border)
        (set-face-background 'child-frame-border "red" my-tooltip-child-frame))
~~~

minad:
And do I understand correctly that it is a real frame going back and forth with the window manager. I am not a big fan of that approach - on my setup this can lead to flicker and I would like to keep the window manager out of the business here. I don't understand why Emacs does not render a rectangle on top of the current frame itself.

PR #21, opened 2021-05-07
Hint desktop-save-mode for child frame

nuance:
By default, the `corfu--frame` child frame is being persisted across sessions saved by `desktop-save-mode`, leading to a random / non-dismissible corfu popup when I launch emacs.

I'm currently using the following advice to prevent this from happening:

```emacs-lisp
  (defun dont-save-frame (orig-fun &rest args)
    "Add the `desktop-dont-save' parameter to the frame."
    (apply orig-fun args)
    (when corfu--frame (set-frame-parameter corfu--frame 'desktop-dont-save t)))
  (advice-add 'corfu--child-frame :around #'dont-save-frame)
```

This PR makes this change more directly.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -255,6 +255,7 @@ Set to nil in order to disable confirmation."
     (set-frame-parameter
      corfu--frame 'background-color
      (face-attribute 'corfu-background :background))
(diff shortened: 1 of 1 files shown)


minad:
Thanks, good find!

PR #43, opened 2021-07-22
Touch up features

jdtsmith:
Ability to summon the popup at any time, even in auto mode, which is a comparative advantage relative to company (which doesn't use the default completion system and so requires rebinding Tab or some such).


minad:
Thanks!

PR #51, opened 2021-08-09
Go back to prompt with scroll on first candidate.

jdtsmith:


minad:
I reconsidered and decided to enable prefix arguments instead as in Vertico. For some reason I didn't implement this yet in Corfu.

Issue #38, opened 2021-07-18
Support non-fixed-width and icon based prefixes


jdtsmith:
Company does a nice job lining up its overlay such that the text in buffer is aligned with the candidates, even when an icon is showing:

<img width="136" alt="image" src="https://user-images.githubusercontent.com/93749/126082661-049e0dc6-27f1-4582-9448-3e4dbdb35f98.png">

Corfu currently aligns the prefix from `:affixation-function` to the left:

<img width="192" alt="image" src="https://user-images.githubusercontent.com/93749/126082679-7303030a-6a1a-49af-adb4-8385faba4f7e.png">

It would be better to move the child frame over so the prefix is to the start of the candidate.  Probably have to compute the maximum prefix width for this offset.  

I also notice that when prefixes have different widths, this makes the left sides of the candidates jagged:

<img width="174" alt="image" src="https://user-images.githubusercontent.com/93749/126083335-9135d398-8d55-4c4e-9b4e-a53de3a7ecc7.png">

Maybe the solution is "don't do that", i.e. either space pad for text prefixes, or use identical widths for icons, etc.  (EDIT: actually scratch that, you are already using specified space to right align suffixes, so no harm in left justifying candidates when prefixes are included, using the same trick).

jdtsmith:
I have a solution for this (that actually uses the method discussed by you and the emacs devs [here](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=47712) to calculate pixel size).  It works quite well even with fractional char width spaces in the prefix:

![image](https://user-images.githubusercontent.com/93749/126092843-3ff35ed1-08b3-4ef4-8eb9-69c254c0d63e.png)

The issue remains that the method used to calculate the maximum _width_:

```elisp
(apply #'max corfu-min-width
                            (mapcar #'string-width lines))
```

cannot account for fractional character width spaces (or images, or invisibility, or ...). So it should probably also use the `window-text-pixel-size` method.  In my tests this was fast enough to run every `show-candidates`.  Let me know if you want a PR.

jdtsmith:
Let me know if you are interested in this.  Here's the utility function I'm calling to find both maximum displayed prefix width and "total" width (though I'll argue below the latter usage is unnecessary):

```elisp
(defun corfu--max-width-pixels (lines)
  "Return the max display width in pixels of LINES."
  (let ((buffer (get-buffer-create " *corfu-width*"))
	(old-buffer (window-buffer)))
[...]

Issue #18, opened 2021-04-29
Option to right-align annotations


fbergroth:
Hi, another happy user here!

Is there any plan for something similiar to `company-tooltip-align-annotations`?

Currently I see:
![image](https://user-images.githubusercontent.com/1211003/116610376-bcc01d80-a935-11eb-9f40-b3dcd44a7bfc.png)

But would prefer to see:
![image](https://user-images.githubusercontent.com/1211003/116610305-aca83e00-a935-11eb-91bc-af8844db5db2.png)

Thanks!

minad:
Please give it a try!

fbergroth:
Looking good!

minad:
See https://github.com/minad/corfu/commit/a56e80e152a555afebe6b8ca1d0963ce4cb93b5f

apoorv569:
How to apply this setting?

minad:
Right aligning is currently not supported.

jdtsmith:
Fairly straightforward to implement by space-padding for fixed space fonts.  It looks possible with mixed pitch.   That would require placing annotation text in the frame’s buffer, measuring pixel sizes, and adjusting specified padding space for alignment.   It can’t be done correctly with overlays so could be a distinctive Corfu advantage.   Thoughts?

minad:
I'd use space :align-to for the alignment and try to avoid performing pixel computations myself in the UI. I've observed before that window-text-pixel-size is slow and unreliable, that's the reason I've avoided it. But please give it a try. If it works fine we can go with that.

minad:
The requirement is reliable and fast. I think for these measurements one has to create a temporary buffer assign it to the frame and measure. I think Emacs 28 also has this documented after I created a bug report.

jdtsmith:
If we try for a `window-text-pixel-size` approach, it’s worth offsetting the frame to line up the prefix to left of candidate text as well, and left-aligning candidates with variable-width prefixes (which could of course include icons).  So basically #38 plus (optional) right alignment of suffix.  

I know you had hoped for some clarity on a new direction in core for affiliation data vs. its presentation.  Not sure if anything has materialized.  If you are still not in favor of “3 item” alignment (even if performant, which I expect it to be from experience), I’ll look at space-padding instead.

minad:
[...]

PR #72, opened 2021-11-07
Implement corfu-align-annotations for both prefix and suffix

jdtsmith:
This implements annotation alignment in corfu, controlled by the custom boolean `corfu-align-annotations`.  Prefixes (if present) are left-padded to the maximum displayed prefix width.  Suffixes (if present) are padded to right-align with the full content width.  Truncation resulting from overflowing the maximum width impacts suffixes first.

All formatting, styling and  alignment (aside from the final bar/margin) has been moved to the function `corfu--format-candidate`, which is now called later, in `corfu--popup-show`.  Until this point, all candidate list items can be strings or `(candidate prefix suffix)` lists.  Note that, when aligning, any pre-existing face information on prefix text is preserved by the prefix-padding, and copied to the surrounding margin padding (margin is placed between prefix and candidate).

This uses simple space-padding for alignment, which means it works correctly for fixed-width content only. Should we desire to support arbitrary variable-pitch fonts/glpyhs/etc. updating  `corfu--format-candidate` and `corfu--max-cand-widths` should be all that is required (presumably using `window-text-pixel-size` in a temporary buffer, or the hidden child frame's window buffer).

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -597,10 +597,10 @@ CAND is either a string or 3 part list of (cand prefix suffix).
 If `corfu-align-annotations' is non-nil, Pad prefix to a total
 width of PREF-WIDTH and right-align the suffix for a total width
 of WIDTH characters (not including prefix), truncating the
-suffix (only) if necessary.  If CURRENT is non-nil, apply the
-face `corfu-current' to the full string.  Insert string MARGIN on
-left and between prefix & candidate, and include RIGHT-MARGIN (if
-non-nil)."
+candidate + suffix if necessary.  If CURRENT is non-nil, apply
+the face `corfu-current' to the full string.  Insert string
+MARGIN on left and between prefix & candidate, and include
+RIGHT-MARGIN (if non-nil)."
   (let ((str
 	 (pcase cand
 	   (`(,cand ,prefix ,suffix)
@@ -609,10 +609,7 @@ non-nil)."
 		  (cw (string-width cand))
 		  (sw (string-width suffix))
 		  (prefix-margin margin)
-		  prefix-pad suffix-pad)
-	      (if (> (+ cw sw) width)
-		  (setq suffix (truncate-string-to-width
-				suffix sw (- (+ sw cw) width))))
+		  prefix-pad suffix-pad candsuff)
 	      (add-face-text-property
 	       0 (length suffix) 'corfu-annotations 'append suffix)
 	      (when corfu-align-annotations
@@ -623,9 +620,13 @@ non-nil)."
 		     0 (length prefix-pad) face 'append prefix-pad)
 		    (setq prefix-margin (copy-sequence prefix-margin))
 		    (add-face-text-property 0 1 face nil prefix-margin)))
+	      (setq candsuff (concat cand suffix-pad suffix))
+	      (if (> (length candsuff) width)
+		  (setq candsuff (truncate-string-to-width candsuff width)))
 	      (apply #'concat
-		     prefix-margin prefix-pad prefix (if (> pref-width 0) prefix-margin)
-		     cand suffix-pad suffix right-margin)))
+		     prefix-margin prefix-pad prefix
+		     (if (> pref-width 0) prefix-margin)
+		     candsuff right-margin)))
 	   (_ (apply #'concat margin cand right-margin)))))
     (if current (add-face-text-property
 		 0 (length str) 'corfu-current 'append str))
--- a/corfu.el
+++ b/corfu.el
@@ -376,6 +376,25 @@ completion began less than that number of seconds ago."
     (set-frame-size corfu--frame width height t)
     (make-frame-visible corfu--frame)))
 
+(defsubst corfu--max-cand-widths (cands)
+  "Return maximum string widths for the given candidates CANDS.
+CANDS is a list of strings, or of lists with elements (candidate
+prefix suffix).  Returns a cons cell 
+
+   (max(prefix) . max(cand+suffix)) 
+
+containing the maximum prefix width and maximum of candiate +
+suffix width, in this order."
+  (let ((widths (mapcar (lambda (cand)
+			  (if (consp cand)
+			      (cons (string-width (cadr cand)) ; prefix
+				    (+ (string-width (car cand))
+				       (string-width (caddr cand))))
+			    (cons 0 (string-width cand)))) ; only candidate
+			cands)))
+    (cons (apply #'max (mapcar 'car widths))
+	  (apply #'max (mapcar 'cdr widths)))))
+
 (defun corfu--popup-show (pos cands &optional curr lo bar)
   "Show CANDS as popup at POS, with CURR highlighted and scrollbar from LO to LO+BAR."
   (let* ((ch (default-line-height))
@@ -632,25 +651,6 @@ RIGHT-MARGIN (if non-nil)."
 		 0 (length str) 'corfu-current 'append str))
     str))
 
-(defsubst corfu--max-cand-widths (cands)
-  "Return maximum string widths for the given candidates CANDS.
-CANDS is a list of strings, or of lists with elements (candidate
-prefix suffix).  Returns a cons cell 
-
-   (max(prefix) . max(cand+suffix)) 
-
-containing the maximum prefix width and maximum of candiate +
-suffix width, in this order."
-  (let ((widths (mapcar (lambda (cand)
-			  (if (consp cand)
-			      (cons (string-width (cadr cand)) ; prefix
-				    (+ (string-width (car cand))
-				       (string-width (caddr cand))))
-			    (cons 0 (string-width cand)))) ; only candidate
-			cands)))
-    (cons (apply #'max (mapcar 'car widths))
-	  (apply #'max (mapcar 'cdr widths)))))
-		     
 (defun corfu--show-candidates (beg end str)
   "Update display given BEG, END and STR."
   (let* ((start (min (max 0 (- corfu--index (/ corfu-count 2)))
--- a/corfu.el
+++ b/corfu.el
@@ -94,6 +94,10 @@ completion began less than that number of seconds ago."
   "Width of the bar in units of the character width."
   :type 'float)
 
+(defcustom corfu-align-annotations t
+  "Whether to align annotations."
+  :type 'boolean)
+
 (defcustom corfu-echo-documentation 0.25
   "Show documentation string in the echo area after that number of seconds."
   :type '(choice boolean float))
@@ -372,8 +376,8 @@ completion began less than that number of seconds ago."
     (set-frame-size corfu--frame width height t)
     (make-frame-visible corfu--frame)))
 
-(defun corfu--popup-show (pos lines &optional curr lo bar)
-  "Show LINES as popup at POS, with CURR highlighted and scrollbar from LO to LO+BAR."
+(defun corfu--popup-show (pos cands &optional curr lo bar)
+  "Show CANDS as popup at POS, with CURR highlighted and scrollbar from LO to LO+BAR."
   (let* ((ch (default-line-height))
          (cw (round (* ch (frame-char-width)) (frame-char-height)))
          (mw (ceiling (* cw corfu-margin-width)))
@@ -383,30 +387,33 @@ completion began less than that number of seconds ago."
          (sbar (concat
                 (propertize " " 'display `(space :width (,(- mw bw))))
                 (propertize " " 'face 'corfu-bar 'display `(space :width (,bw)))))
-         (width (min corfu-max-width
+	 (mcw (corfu--max-cand-widths cands))
+	 (max-pref-width (car mcw))
+	 (max-cand-width (cdr mcw))
+	 (width (min corfu-max-width
                      (frame-width)
-                     (apply #'max corfu-min-width
-                            (mapcar #'string-width lines))))
-         (row 0)
+		     (max corfu-min-width
+			  max-cand-width)))
          (pos (posn-x-y (posn-at-point pos)))
          (x (or (car pos) 0))
-         (y (or (cdr pos) 0)))
+         (y (or (cdr pos) 0))
+	 (bar-end (if lo (+ lo bar)))
+	 (x-offset (+ (* mw (if (> max-pref-width 0) 2 1))
+		      (* max-pref-width cw)))
+	 (lines
+	  (cl-loop for cand in cands and row from 0
+		   for right-margin = (if (and lo (<= lo row bar-end))
+					  sbar margin)
+		   collect (corfu--format-candidate
+			    cand max-pref-width width (eq row curr)
+			    margin align right-margin))))
+    ;; per-line layout: M(ppPPM)?CCCCCCC(pppp)?SSSSS(M|B)
+    ;; M=margin, p=padding, P=prefix, C=candidate, S=suffix, B=bar
     (corfu--make-frame
-     (- x mw) y
-     (+ (* width cw) mw mw) (* (length lines) ch)
-     (mapconcat (lambda (line)
-                  (let ((str (concat
-                              margin
-                              (truncate-string-to-width line width)
-                              align
-                              (if (and lo (<= lo row (+ lo bar)))
-                                  sbar margin))))
-                    (when (eq row curr)
-                      (add-face-text-property
-                       0 (length str) 'corfu-current 'append str))
-                    (setq row (1+ row))
-                    str))
-                lines "\n"))))
+     (- x x-offset) y
+     (+ (* (+ max-pref-width width) cw) (* mw (if (> max-pref-width 0) 3 2)))
+     (* (length lines) ch)
+     (string-join lines "\n"))))
 
 (defun corfu--popup-hide ()
   "Hide Corfu popup."
@@ -566,21 +573,14 @@ completion began less than that number of seconds ago."
   (completion-in-region-mode -1))
 
 (defun corfu--affixate (candidates)
-  "Annotate CANDIDATES with annotation function."
+  "Annotate CANDIDATES with affixation or annotation function."
   (if-let (aff (or (corfu--metadata-get corfu--metadata 'affixation-function)
                    (plist-get corfu--extra :affixation-function)))
       (funcall aff candidates)
     (if-let (ann (or (corfu--metadata-get corfu--metadata 'annotation-function)
                      (plist-get corfu--extra :annotation-function)))
-        (mapcar (lambda (cand)
-                  (let ((suffix (or (funcall ann cand) "")))
-                    (list cand ""
-                          ;; The default completion UI adds the `completions-annotations' face
-                          ;; if no other faces are present. We use a custom `corfu-annotations'
-                          ;; face to allow further styling which fits better for popups.
-                          (if (text-property-not-all 0 (length suffix) 'face nil suffix)
-                              suffix
-                            (propertize suffix 'face 'corfu-annotations)))))
+        (mapcar (lambda (cand) ; suffix only
+                  (list cand "" (or (funcall ann cand) "")))
                 candidates)
       candidates)))
 
@@ -590,14 +590,66 @@ completion began less than that number of seconds ago."
   "Return PROP from METADATA."
   (cdr (assq prop metadata)))
 
-(defun corfu--format-candidate (cand)
-  "Format annotated CAND string."
-  (replace-regexp-in-string
-   "[ \t]*\n[ \t]*" " "
-   (if (consp cand)
-       (concat (cadr cand) (car cand) (caddr cand))
-     cand)))
+(defun corfu--format-candidate (cand pref-width width &optional current margin
+				     &rest right-margin)
+  "Format candidate, returning a string with styling and alignment.
+CAND is either a string or 3 part list of (cand prefix suffix).
+If `corfu-align-annotations' is non-nil, Pad prefix to a total
+width of PREF-WIDTH and right-align the suffix for a total width
+of WIDTH characters (not including prefix), truncating the
+suffix (only) if necessary.  If CURRENT is non-nil, apply the
+face `corfu-current' to the full string.  Insert string MARGIN on
+left and between prefix & candidate, and include RIGHT-MARGIN (if
+non-nil)."
+  (let ((str
+	 (pcase cand
+	   (`(,cand ,prefix ,suffix)
+	    (setq cand (replace-regexp-in-string "[ \t]*\n[ \t]*" " " cand))
+	    (let ((pw (string-width prefix))
+		  (cw (string-width cand))
+		  (sw (string-width suffix))
+		  (prefix-margin margin)
+		  prefix-pad suffix-pad)
+	      (if (> (+ cw sw) width)
+		  (setq suffix (truncate-string-to-width
+				suffix sw (- (+ sw cw) width))))
+	      (add-face-text-property
+	       0 (length suffix) 'corfu-annotations 'append suffix)
+	      (when corfu-align-annotations
+		  (setq prefix-pad (make-string (max 0 (- pref-width pw)) ? )
+			suffix-pad (make-string (max 0 (- width cw sw)) ? ))
+		  (when-let ((face (get-text-property 0 'face prefix)))
+		    (add-face-text-property
+		     0 (length prefix-pad) face 'append prefix-pad)
+		    (setq prefix-margin (copy-sequence prefix-margin))
+		    (add-face-text-property 0 1 face nil prefix-margin)))
+	      (apply #'concat
+		     prefix-margin prefix-pad prefix (if (> pref-width 0) prefix-margin)
+		     cand suffix-pad suffix right-margin)))
+	   (_ (apply #'concat margin cand right-margin)))))
+    (if current (add-face-text-property
+		 0 (length str) 'corfu-current 'append str))
+    str))
 
+(defsubst corfu--max-cand-widths (cands)
+  "Return maximum string widths for the given candidates CANDS.
+CANDS is a list of strings, or of lists with elements (candidate
+prefix suffix).  Returns a cons cell 
+
+   (max(prefix) . max(cand+suffix)) 
+
+containing the maximum prefix width and maximum of candiate +
+suffix width, in this order."
+  (let ((widths (mapcar (lambda (cand)
+			  (if (consp cand)
+			      (cons (string-width (cadr cand)) ; prefix
+				    (+ (string-width (car cand))
+				       (string-width (caddr cand))))
+			    (cons 0 (string-width cand)))) ; only candidate
+			cands)))
+    (cons (apply #'max (mapcar 'car widths))
+	  (apply #'max (mapcar 'cdr widths)))))
+		     
 (defun corfu--show-candidates (beg end str)
   "Update display given BEG, END and STR."
   (let* ((start (min (max 0 (- corfu--index (/ corfu-count 2)))
@@ -607,7 +659,7 @@ completion began less than that number of seconds ago."
          (bar (ceiling (* corfu-count corfu-count) corfu--total))
          (lo (min (- corfu-count bar 1) (floor (* corfu-count start) corfu--total)))
          (cands (funcall corfu--highlight (seq-subseq corfu--candidates start last)))
-         (ann-cands (mapcar #'corfu--format-candidate (corfu--affixate cands))))
+         (ann-cands (corfu--affixate cands)))
     ;; Nonlinearity at the end and the beginning
     (when (/= start 0)
       (setq lo (max 1 lo)))


jdtsmith:
As an example, here is the resulting alignment using a custom affixation function which (if no other affixation function is set) interprets the `:company-kind` property, as provided here by `lsp-mode`:

image

In this case the kind and suffix data are redundant, but this wouldn't always be the case.

corfu.el
```diff
@@ -94,6 +94,10 @@ completion began less than that number of seconds ago."
   "Width of the bar in units of the character width."
   :type 'float)
 
+(defcustom corfu-align-annotations t
```
minad, line 97: Do we need this customization option? Why not align always?

corfu.el
```diff
@@ -383,30 +387,33 @@ completion began less than that number of seconds ago."
          (sbar (concat
                 (propertize " " 'display `(space :width (,(- mw bw))))
                 (propertize " " 'face 'corfu-bar 'display `(space :width (,bw)))))
-         (width (min corfu-max-width
+	 (mcw (corfu--max-cand-widths cands))
```
minad, line 390: Is there some whitespace issue here?

corfu.el
```diff
@@ -566,21 +573,14 @@ completion began less than that number of seconds ago."
...
                      (plist-get corfu--extra :annotation-function)))
-        (mapcar (lambda (cand)
-                  (let ((suffix (or (funcall ann cand) "")))
-                    (list cand ""
-                          ;; The default completion UI adds the `completions-annotations' face
-                          ;; if no other faces are present. We use a custom `corfu-annotations'
-                          ;; face to allow further styling which fits better for popups.
-                          (if (text-property-not-all 0 (length suffix) 'face nil suffix)
-                              suffix
-                            (propertize suffix 'face 'corfu-annotations)))))
```
minad, line 583: So the idea is to always append the corfu-annotations face now? Is this a good idea? It is a deviation from how the minibuffer ui works, but maybe reasonable for the popup.

minad:
> All formatting, styling and alignment (aside from the final bar/margin) has been moved to the function corfu--format-candidate, which is now called later, in corfu--popup-show. Until this point, all candidate list items can be strings or (candidate prefix suffix) lists. Note that, when aligning, any pre-existing face information on prefix text is preserved by the prefix-padding, and copied to the surrounding margin padding (margin is placed between prefix and candidate).

Hmm, this is a refactoring I dislike. I intentionally had the popup code decoupled from the candidate formatting code.

corfu.el
jdtsmith: Since backends are apparently free to put whatever they want into affixations (including alignment spaces!) it seems sensible to allow people to turn it off if corfu alignment is interfering.

corfu.el
jdtsmith: We could check for a nil face and not apply in that case.

corfu.el
jdtsmith: > > All formatting, styling and alignment (aside from the final bar/margin) has been moved to the function corfu--format-candidate, which is now called later, in corfu--popup-show. Until this point, all candidate list items can be strings or (candidate prefix suffix) lists. Note that, when aligning, any pre-existing face information on prefix text is preserved by the prefix-padding, and copied to the surrounding margin padding (margin is placed between prefix and candidate).
> 
> Hmm, this is a refactoring I dislike. I intentionally had the popup code decoupled from the candidate formatting code. If possible I would like to keep the layering of the different concerns.

It will be challenging to keep your original setup, since you now must compute the widths of the candidates + suffixes quite late, so as to be able to set the width correctly.  Take a look, but I view this as a simplification, since in the current setup, the "formatting" (alignment, styles, etc.) are now self-contained, instead of being spread out over several different functions.

corfu.el
minad: Yes, this would be better.

jdtsmith:
> Hmm, this is a refactoring I dislike. I intentionally had the popup code decoupled from the candidate formatting code. If possible I would like to keep the layering of the different concerns.

It will be challenging to keep your original setup, since you now must compute the widths of the candidates + suffixes quite late, so as to be able to set the width correctly. Take a look, but I view this as a simplification, since in the current setup, the "formatting" (alignment, styles, etc.) are now all self-contained in a single function, instead of being spread out over several different functions.  I.e. even less of the formatting code is now in `corfu--popup-show`.

minad:
> we'll have to pass into it the result of (corfu--max-cand-widths cands) as another parameter. 

Yes, I would prefer that.

However there is another issue with your patch - you effectively removed the line truncation. corfu-max-width is not respected anymore if I understand correctly?

corfu.el
jdtsmith: This is already the case (i.e. `'append` is set). So if some backend sets the text foreground, that will take precedence. I still think it's bad idea to let backends be in charge of styling though.

corfu.el
jdtsmith: I.e. they could make this variable buffer-local and disable alignment in certain "problem" modes with CAPFs that do their own alignment in an affixation/annotation function.

jdtsmith:
> However there is another issue with your patch - you effectively removed the line truncation. corfu-max-width is not respected anymore if I understand correctly?

No, `corfu--format-candidate` is in charge of truncation.  It truncates the suffix if the total candidate + suffix width exceeds the pre-computed width.  If that is insufficient, the buffer width will simply leave part of the candidate text off the frame.  Though now that I say that, did you truncate primarily to prevent scrollbars appearing in the child frame (not even sure that's possible)?

minad:
> If that is insufficient, the buffer width will simply leave part of the candidate text off the frame. Though now that I say that, did you truncate primarily to prevent scrollbars appearing in the child frame (not even sure that's possible)?

Yes, unfortunately this is insufficient. This pushes our own scrollbar (the thin bar at the right) outside the popup :(

jdtsmith:
> > we'll have to pass into it the result of (corfu--max-cand-widths cands) as another parameter.
> 
> Yes, I would prefer that. But only the prefix width is needed to move the popup to the left.

But you also need max "candidate + suffix" width to size the frame, given the non-character-width internal alignment being inserted between prefix and candidate (which `string-width` is oblivious to).  We can obviously pass those both in together.

But there are a few other issues. The original `popup-show` formats margins and bar (including right-aligning the right one), computes frame width, truncates lines, and applies the current face on the selected row.  Since the "format" now includes optional internal margins (between prefix and candidate), and altering left margin BG (for prefixes with background color), calling the formatter prior to `popup-show` as you hope for will require the earlier calling function to setup at least the margins, but likely also scrollbars, etc.  This will basically gut `popup-show` and make it simply a thin wrapper for `corfu-make-frame`.

jdtsmith:
> > If that is insufficient, the buffer width will simply leave part of the candidate text off the frame. Though now that I say that, did you truncate primarily to prevent scrollbars appearing in the child frame (not even sure that's possible)?
> 
> Yes, unfortunately this is insufficient. This pushes our own scrollbar (the thin bar at the right) outside the popup :(

Ah right.  In that case I can first glue prefix + suffix together with any padding, then trim them to the respect the limit together. Note that I'm not counting prefix against the max-width.  But we need to first decide what to do with `popup-show` (see above).

minad:
I would like to simplify the discussion a bit and take smaller step. I experiment a bit and found multiple issues with my existing code. In particular frame resizing does not work correctly on my system (not pixelwise), due you observe this too?

I think it is easier if we proceed as I did here https://github.com/minad/corfu/commit/1c7fdb15c200cee9c6836dfc861f7c5216f606e9, essentially keeping the original layering of the functions intact.

corfu.el
minad: Are there really backends which add alignment spaces? Note that my current code assumes that no alignment spaces are present. We are doing `string-width` computations all over the place. I would like to proceed for now without the customization option and figure out a simpler solution first. We can easily add this back - and then you are right that one could set this buffer locally or somehow even per-capf.

corfu.el
minad: In this case the faces will be mixed. But this is also a secondary concern which is not related to how we fix the alignment. The current code matches Vertico but I am open to changing this to always overriding with the corfu-annotations face, under the assumption that Corfu is somehow a special frontend and is in charge of styling.

minad: We could always apply `substring-no-properties` to the returned strings in case we find that backends mess up the careful Corfu formatting too much ;)

jdtsmith:
> I would like to simplify the discussion a bit and take smaller steps. I experiment a bit and found multiple issues with my existing code. In particular frame resizing does not work correctly on my system (not pixelwise), do you observe this too?

Which frame resizing?  The full parent frame?  That has no impact on the child frame here (MacOS).  I doubt very much this PR would introduce new frame resizing bugs since it's simply adding spaces for padding (+ styling).  I could certainly see pixel-based resizing might do so.

> I think it is easier if we proceed as I did here [1c7fdb1](https://github.com/minad/corfu/commit/1c7fdb15c200cee9c6836dfc861f7c5216f606e9), essentially keeping the original layering of the functions intact. In the next step the popup should be shifted to the left in order to account for the prefix. 

Not sure I'm following what is different about this approach, can you expand?  The only real difference in flow here from before is that the current functionally of `corfu--format-candidate` was formerly spread across multiple functions, and called earlier, in `show-candidates`.  Here I've consolidated the padding, text property setting/etc., margin wrapping, etc. to a single function, and moved it later, which in my view is a simplification (but you may not agree). 

> Furthermore we could also implement pixel-based alignment. But then we should first fix the pixelwise resizing of the frame. And the frame should also take a width argument in pixels, since the current popup width computation based on string-width is invalid.

I'm not sure I understand what you mean by pixel-wise resizing.  How does one resize a child frame pixel-wise?  Once my child frames are born, they stay the same size until destroyed which means they can fall outside the parent frame:

image

I do think pixel-based alignment would be a distinctive addition (better than overlay-based UI's) and is certainly possible from what I can see.  But it comes with a whole set of decisions that need to be made, including whether to use a hidden custom pixel-size-measurement buffer or the actual child-frames window buffer (which means deferring formatting until quite late). It's also unclear whether emacs will eventually provide some separation of data and presentation in the CAPF specification.

minad:
> Which frame resizing?

On my system the child frame is always resized in steps of char-width instead of pixelwise. This leads to large margins on the right. I have not yet figured out how to fix this.

> Not sure I'm following what is different about this approach, can you expand? The only real difference in flow here from before is that the current functionally of corfu--format-candidate was formerly spread across multiple functions, and called earlier, in show-candidates.

The problem is that we have two formattings going on in my current design: candidate formatting and popup formatting (margin scrollbar). These formattings are done separately now in 1c7fdb15c200cee9c6836dfc861f7c5216f606e9. The function `corfu--format-candidates` is completely responsible for the candidates while the popup does only the scrollbar. The popup also does not truncate the candidate strings, which it did before and which was not that good. The code in its current form is at least to a certain extent more reusable, since the popup code does not call the candidate formatting.

> I do think pixel-based alignment would be a distinctive addition (better than overlay-based UI's) and is certainly possible from what I can see. 

Indeed.

> But it comes with a whole set of decisions that need to be made, including whether to use a hidden custom pixel-size-measurement buffer or the actual child-frames window buffer (which means deferring formatting until quite late).

I am not sure I want to push the pixel computations to the very end. The child frame code is notoriously problematic and full of hacks. If we have to wait until the child frame appeared or just before it is made visible to make the formatting, we will only get more problems. I think it is better to perform the pixel calculations earlier and then just display the end result.

jdtsmith:
> > Which frame resizing?
> 
> On my system the child frame is always resized in steps of char-width instead of pixelwise. This leads to large margins on the right. I have not yet figured out how to fix this.

Interesting.  So it rounds up to the next multiple of charsize?  Surprising.  I don't see that here.   Does `frame-pixel-width` at least report correctly after creation?

> > Not sure I'm following what is different about this approach, can you expand? The only real difference in flow here from before is that the current functionally of corfu--format-candidate was formerly spread across multiple functions, and called earlier, in show-candidates.
> 
> The problem is that we have two formattings going on in my current design: candidate formatting and popup formatting (margin scrollbar). These formattings are done separately now in [1c7fdb1](https://github.com/minad/corfu/commit/1c7fdb15c200cee9c6836dfc861f7c5216f606e9). The function `corfu--format-candidates` is completely responsible for the candidates while the popup does only the scrollbar. The popup also does not truncate the candidate strings, which it did before and which was not that good. The code in its current form is at least to a certain extent more reusable, since the popup code does not call the candidate formatting function.

It's a reasonable approach.  But there is some coupling between these two once you introduce alignment and (e.g.) left margin recoloring.  BTW, with pixel-alignment and variable width entities, you can no longer rely on simple `truncate-string-to-width` to avoid blowing the width.  Looking ahead to pixel sizing is why I'm enforcing a max width only on candidate+suffix (which hopefully `string-width` can give correct results for).  But if suffix is in a variable width font, all bets are off. 

> > I do think pixel-based alignment would be a distinctive addition (better than overlay-based UI's) and is certainly possible from what I can see.
> 
> Indeed.
> 
> > But it comes with a whole set of decisions that need to be made, including whether to use a hidden custom pixel-size-measurement buffer or the actual child-frames window buffer (which means deferring formatting until quite late).
> 
> I am not sure I want to push the pixel computations to the very end. The child frame code is notoriously problematic and full of hacks. If we have to wait until the child frame appeared or just before it is made visible to make the formatting, we will only get more problems. I think it is better to perform the pixel calculations earlier and then just display the end result.

In an earlier incarnation of prefix alignment from #38  (when I was using icon prefixes) I had:

```elisp
(defun corfu--max-width-pixels (lines)
  "Return the max display width in pixels of LINES."
  (let ((buffer (get-buffer-create " *corfu-width*"))
	(old-buffer (window-buffer)))
    (with-current-buffer buffer ; see discussion of bug #47712
      (erase-buffer)
      (insert lines)
      (set-window-buffer (selected-window) buffer)
      (prog1
	  (car (window-text-pixel-size nil (point-min) (point-max)))
 	(set-window-buffer (selected-window) old-buffer)))))
```

Is something like this what you mean?  That could happen any time, well in advance of child frame availability.  

`<BRAINSTORMING>` For a full pixel-based approach (arbitrary variable-width fonts, pixel-sized spaces, arbitrary width and height glyphs, etc.), the procedure would look something like:

1. Style and insert all the prefixes into the measurement buffer, measure per-line pixel width and its maximum.  
2. Pad each prefix with specified space to make them all identical width in pixels.
2. Erase and insert all the candidates into the measurement buffer.   Measure maximum and per-candidate pixel widths. 
3. Erase and insert all the (styled) suffixes into the measurement buffer.  Measure maximum and per-suffix pixel widths.
1. Compute a maximum pixel width for the child frame based on margin, above-measured candidate + suffix pixel widths, and the maximum width _in pixels_ the user wants to accommodate for the popup frame. 
5. If candidate and suffix together are too wide in pixels, drop suffix and candidate characters (one at a time?) until together they are no longer too wide.   Possibly can use average width per character to do this faster. 
6. If a suffix has survived the cut, concat candidate and suffix with padding between sufficient to precisely right-align the suffix.  
7. Concat margin, prefix, prefix padding, candidate+suffix, margin|bar.
8. Insert _the entire set of lines_ in the hidden measurement buffer and measure the maximum pixel _height_ of the lines.  
9. Erase and insert these lines in the child frame's presentation buffer. 
8. Create the frame of the correct pixel width (possibly dilating margins to "spread the delta" if, as in your system the frame doesn't give you that exact width back), and the measured pixel height (not much to do about that one).  
9. Assign the buffer to the child frame's window. 

I think this will all be fast enough for (typical) 10-20 entries, but for scrolling performance, it would also probably be worth caching and reusing pixel size info in a text-property on candidates as they appear, so that only the latest unseen prefix/canddiate/suffix pixel widths need to be measured.  

So... pretty heavy duty.  Probably the reason I reached for space-padding (as company does) first.

minad:
> Interesting. So it rounds up to the next multiple of charsize? Surprising. I don't see that here. Does frame-pixel-width at least report correctly after creation?

I have to investigate. It could be due to the window manager or some other configuration. On your screenshot it does not look as if this happens for you?

> It's a reasonable approach. But there is some coupling between these two once you introduce alignment and (e.g.) left margin recoloring. BTW, with pixel-alignment and variable width entities, you can no longer rely on simple truncate-string-to-width to avoid blowing the width.

You are right about that. The idea is to pass an additional pixel WIDTH argument to `corfu--show-popup`. Then `corfu--format-candidates` should format the candidates such that it fits within this pixel width and not use `string-width`/`truncate-string-to-width`. Furthermore the popup should be shifted to the left by the prefix width in pixels which can also be computed by `corfu--format-candidates`. This way we can upgrade to pixelwise alignment.

minad:
> BRAINSTORMING

Yes this is how it could be implemented. As you said, we probably need caching. But note that we are not allowed to modify the candidate strings, so the cache must be separate. I also dislike how complicated this is. Therefore I went with char alignment in Corfu for now.

I would first look around a bit, I've seen a project which tried to align org tables in the presence of variable pitch characters. It would be good to check first how they are solving the problem.

corfu.el
jdtsmith: I don't know of any, but it's technically perfectly allowed (if foolish).  LSP-mode inserts a leading character in the annotation string, which I find strange.

jdtsmith: I'm of course happy to make it the default if you want to cut down on customize-clutter.

jdtsmith:
> I have to investigate. It could be due to the window manager or some other configuration. On your screenshot it does not look as if this happens for you?

I don't ever see any extra width.  The extra space before the annotation is actually inserted by LSP-mode.

> > It's a reasonable approach. But there is some coupling between these two once you introduce alignment and (e.g.) left margin recoloring. BTW, with pixel-alignment and variable width entities, you can no longer rely on simple truncate-string-to-width to avoid blowing the width.
> 
> You are right about that. The idea is to pass an additional pixel WIDTH argument to `corfu--show-popup`. Then `corfu--format-candidates` should format the candidates such that it fits within this pixel width and not use `string-width`/`truncate-string-to-width`. Furthermore the popup should be shifted to the left by the prefix width in pixels which can also be computed by `corfu--format-candidates`. This way we can upgrade to pixelwise alignment.

Even with fixed-width chars, you need to pass in two widths at present: prefix width and candidate + suffix max width.  The reason is that internal partial width spaces are now included, which `string-width` cannot understand.  A pixel-width measurement inside `show-popup` could account for this indeed (though it still has the problem of "how many characters do I trim to come back within the pixel width cap?").  But the real question: how far to go down the pixel-width rabbit hole?  What if completion suffix info is styled with a variable width font, as is common in other IDE's?  It's fairly elaborate to do this right I think (see above).   If it got to the point of working well, this would be a major distinctive advantage of corfu: use whatever styling you want, _it just works_.

jdtsmith:
Oh I see you actually pushed a commit to main?  I tried it and it works OK, not great.  Unfortunately, my lovely new lsp kind method looks like this now:

image

Missing:

- true right-alignment of suffixes, as opposed to left alignment to a "far enough" position.
- inherited background color for the prefix padding.
- re-coloring left margin background based on prefix. 
- right-aligned prefixes. 
- face styling on the suffixes of affixation-functions.  

I'm not sure how to proceed here; lots of work went into the PR and I wouldn't want it to go to waste.

minad:
I would like to proceed here and not let this go to waste. However you got carried away a bit with too many unrelated steps, which I cannot check and confirm independently since they essentially work only on your setup. I don't have access to the kind/affixation function you used. For testing purposes it would be good to have a demo function for example for elisp mode which makes it the easiest to test against emacs -Q. The list of things you listed above are easily fixable one by one in separate commits. Then as I mentioned on my setup the child frame doesn't even work yet properly. This is not related to your PR of course but this prevents me from checking if the pixel alignment works properly.

> true right-alignment of suffixes, as opposed to left alignment to a "far enough" position.

Agree. This is easy to fix in corfu--format-candidates.

> inherited background color for the prefix padding.

Looks nice, but are there affixation functions which actually provide such prefixes?

> re-coloring left margin background based on prefix.

Okay, this one I missed entirely. You are right that it looks nicer with this change and this is certainly a reason to mix formatting and the popup code as you did.

> right-aligned prefixes.

Why is this desired?

> <full space padding between prefix and candidate.

We could add a half space or only a margin. But then the separator should be configurable with a separate variable.

> face styling on the suffixes of affixation-functions.

Which I argued should potentially not be done. The affixation function by definition may apply custom styling. But we could as well decide to overwrite it as you did. But it is a separate change.

minad:
I use this affixation function for testing now:

~~~ elisp
(advice-add #'elisp-completion-at-point :filter-return #'corfu-elisp-affixation-advice)

(defun corfu-elisp-affixation-advice (result)
  (append result (list :affixation-function #'corfu-elisp-affixation)))

(defun corfu-elisp-affixation (cands)
  (mapcar (lambda (cand)
            (list cand
                  (propertize
                   (cond
                    ((commandp (intern-soft cand)) "com")
                    ((fboundp (intern-soft cand)) "fu")
                    (t "v"))
                   'face 'corfu-annotations)
                  (propertize
                   (cond
                    ((commandp (intern-soft cand)) "Command")
                    ((fboundp (intern-soft cand)) "Function")
                    (t "Variable"))
                   'face 'corfu-annotations)))
          cands))
~~~

minad:
I finally found the issue of the wrong frame resizing - the calculation of the character width was wrong. Fixed in https://github.com/minad/corfu/commit/69407ed50dc51ab6d64a885ffca510ea92da4012.

jdtsmith:
[Here](https://github.com/jdtsmith/corfu-kind-prefix) is the small package I wrote to map `:company-kind` into an affixation function (if none set).  OK either way on prefix alignment.  I vote for styling affixation suffixes the same as annotation: I expect most users won't know or care where their suffixes came from and would be surprised if they weren't styled the same.   Will leave this in your hands.

minad:
> Here is the small package I wrote to map :company-kind into an affixation function (if none set).

Cool, this looks interesting. I wonder why it depends on Corfu though. Since you adjust the affixation function here, which is understood by Corfu you could instead adjust the capf wrapper with an advice, such that it is added there.

> I vote for styling affixation suffixes the same as annotation: I expect most users won't know or care where their suffixes came from and would be surprised if they weren't styled the same.

I lean to not doing this, since it is also not applied for completing-read. The affixation function does the styling itself. It is not about the user - it only affects the author of the affixation-function.

jdtsmith:
Couldn’t think of a safe place to shim in a kind-translator affixation function.  Ideas?  I can also add suffix highlighting in this package as well.  It would also be easy to add icon support to this translator package; in fact I have an early prototype that used the vscode icons.  Either they have to be one char wide (svg?) or it would have to do pixel width calc.  That part is fast and reliable with the function I showed above.

minad:
@jdtsmith It would be nice to have an icon package first, which could be used by both Corfu and Company. See my comment https://github.com/minad/corfu/issues/18#issuecomment-963211338.

minad:
I just saw your kind-prefix package. I am looking forward to this. But you may want to choose a more meaningful name?

There you overwrite the completion-in-regiom-function which is of course one possibility but it seems a bit brute force. This will lead to initialisation order issues with Corfu and consult-completion-in-region. Maybe we can figure out a better way. 

Then the question is how we proceed here. As I see it you need at least the following:

1. The popup should move to the right by the width of the prefix (for now in fixed char widths?).
2. The margin on the left should inherit the face of the prefix.
3. The padding between prefix and candidate should be configurable or even removed? For this I am not sure on the best solution, since this depends highly on the affixation function being used. Maybe the affixation function is not the right tool for this an one needs an additional corfu-prefix-function.

minad:
On the other hand the affixation function may never get implemented by any completion backend due to the refusal of company to implement it. This means we can reinterpret and reuse it in our own ways. For example not adding a padding after the prefix and assuming that the affixation function takes care of that. But then we could as well drop affixation support from Corfu and implement our own corfu-prefix-function extension instead.

minad:
Another option - maybe the best so far would be to add support for company-kind here directly. But not really implementing support - only adding a corfu-kind-formatter-function variable which can be set from outside, e.g., by your package.

jdtsmith:
OK I have generalized my package to [kind-prefix](https://github.com/jdtsmith/kind-prefix) and added icons.  It wraps the `completion-in-region-function` to add a custom affixation function which takes its prefix from `:company-kind`.  For icons, I'm using [`svg-lib`](https://github.com/rougier/svg-lib), which taps into the huge free "material" SVG icon library.  It's _very_ simple to specify icons: just a string name, which gets downloaded and cached locally and drawn at the appropriate size.  This makes it utterly trivial for users to [search](https://materialdesignicons.com) and customize personalized icons that work for them.  The other big advantage of `svg-lib` is it draws icons to precisely occupy the space of 2x1 characters, so you can just use a display function on two characters and everything lines up perfectly.  No pixel sizing at all required. 

I've tested it with my branch of corfu. It looks really good:

image

Once the corfu dependency was removed, I did need to let the package know about the background face (e.g. `corfu--background`).  This is because the affixation function is called _before_ the popup frame is created, so the frame default background is incorrect.  Not too big of a problem.  

I tried the latest 0.14 release of corfu, and found it does a pretty good job with kind-prefix.  Biggest things missing:

- (Simple) Inheriting the background color of the prefix into the adjacent thin spacers. 
- Offsetting the prefix to line up with buffer text.
- Highlighting the prefix on the currently selected candidate line.

For the offsetting, I'd advocate the simplest possible solution: "max character width" prefix offsetting.  This will work perfectly for text prefixes, and `svg-lib` proves it is quite possible to constrain icons to fit the character spacing and look good at different sizes.  Here's a view at 18 point (again just 2x1 characters):

image

For the prefix-current-highlighting, some means of altering the prefix BG color would be coolest.  A possibly cheap and easy fix is as follows: remove the foreground specification of `corfu--background` and then set it as priority (no `'append`), so the thin prefix spacers become current-colored:

image

Obviously would have to move the spacer bar background to an even higher priority.  BTW, when testing this I found that `frame-char-width` doesn't update correctly  on my system as the buffer font is changed.  See #74.

minad:
@jdtsmith I already answered in the past.

minad:
I added a commit for the prefix movement (https://github.com/minad/corfu/commit/00e402e301abc1edf7ab1a98ef439d2f552301a7) of the popup.

There are still these open issues left for good icon support:

1. Inheriting the background face for the margin
2. Figuring out on what to do with the padding. See what I wrote above in https://github.com/minad/corfu/pull/72#issuecomment-964639925. A corfu-kind-formatter-function is probably the way to go. This would also simplify things on your end since the user would only have to set this variable to your kind-badge function.
3. The prefix highlighting for the current line (very minor issue, maybe I even prefer to not highlight it)

jdtsmith:
>  I already answered in the past ;)

You have some impressive time traveling skills!  

Name suggestions?  It was corfu-kind-prefix but I took the corfu out, and just lopped it off.  Happy to take any ideas.  `kind-badges`?  I'm using "badge" to mean either an icon or a short text snippet.  `badgers`?  I always liked badgers.

First a note on philosophy: an affixation or annotation function in my view should never deal in alignment, spacing, padding, etc.  That is a UI's job.  Different UI's may take different approaches.  The fact that I'm adding color in an affixation already feels... weird.  But I draw the line at that ;).  Backends should supply _data_, front-ends should format it up nice for the user.  Someday somebody might want some crazy pie-chart wheel interface or something.  I give data (with a bit of styling), you work out how to lay it out.  That's why it annoys me that lap-mod pre-pads its annotation data with a space.  

In terms of hooking in, formerly I was advising corfu-setup, and I struggled to find a good location to insert an affixation.  I don't clobber the completion-in-region-function, I just wrap it, so it still gets called as-is (after setting up the affixation function).  Also I had to pass in to the closure the annotation metadata right away, because by the time the affixation function gets called, it's not accessible (or at least I couldn't figure out how to get to it).  But I'm _definitely_ keen to hear of any less intrusive ideas you may have for shimming in the :company-kind -> affixation translator. 

I'm also happy to abandon the affixation sleight of hand altogether if corfu provides an external hook for kind.  But that might tie it unnecessarily back to corfu (in which case maybe corfu returns to the name?).  Are there other completion-in-region UI's you know of that support affixation?  I know company doesn't (and of course doesn't need it for this case).  But maybe others will appear?  In principles it's a nice generic place to plug into a UI — a layer you can insert between backend and front-end for "setting up 3 parts".   And since it's part of the Emacs 28 standard, perhaps others (company-aside) will support it in the future. 

If they do, the awkward thing will then be if lots of backends eventually start supporting affixation themselves, it would render my translator inoperative/pointless.  For a brief while I was calling pre-existing affixation-functions and just overwriting their prefixes.  But I thought better of that.  But maybe if backends are doing the right thing by then, it won't matter, and this can be retired.  Again, since most things are focused on company, affixations probably won't catch on.  Which means we can just use them for our own purposes and see who joins us. 

For the padding between prefix and candidate: I suspect users may want to configure this (up to and including removing it).  Inheriting the colors for this padding/margin is good too, and looks great as you can see, but I'm still searching for a good current-highlighting solution for the prefix.  This likely has to be the UI's job, since a translator shim doesn't know anything about the current-selection etc.  

If Emacs had transparency, you could have a stack of background colors like BG-1 (50%) BG-2 (50%) BG-3 (0%). It's not hard to simulate this (see `kind-prefix--rgb-blend`).    Maybe when the colors are added to the prefix padding, a "related current BG color" could be computed and saved at the back end of the face list?  Or something simpler...

minad:
A very long answer - I try to be brief and explain my problem a bit better:

Corfu inserts spaces between the components: prefix space candidate space suffix. For your icons you don't want to have a space: prefix candidate space suffix (correct?), in order to give you more freedom. Now the problem is that for the affixation functions I always should insert spaces since I cannot assume that the prefix is indeed an icon which does not require such a padding. Therefore I may need an additional entry point only for icons. There will be no hooking back to Corfu. The user can set corfu-kind-formatter-function=kind-badge.

The kind-badge function can probably be used by Company too, even if it does not support affixation-functions.

minad:
As for the name it may be better to have icons or badge in the name. Badge does not sound very familiar, so a better name would be kind-icons.

jdtsmith:
> A very long answer - I try to be brief and explain my problem a bit better:
> 
> Corfu inserts spaces between the components: prefix space candidate space suffix. For your icons you don't want to have a space: prefix candidate space suffix (correct?), in order to give you more freedom. 

I'd suggest `margin prefix thinspace candidate thinspace suffix margin|bar` for everything, text/icon/whatever.  Where `thinspace` has a configurable width down to 0 (maybe even two configuration settings).  That way people can tune it how they want.  As you can see above (on the badly highlighted current line), I do have margin before and after the icon, and it looks good to me.  I'd probably just leave it in even if I had the option to remove it. 

> Now the problem is that for the affixation functions I always should insert spaces since I cannot assume that the prefix is indeed an icon which does not require such a padding. Therefore I may need an additional entry point only for icons. There will be no hooking back to Corfu. The user can set corfu-kind-formatter-function=kind-badge.

Well that's a form of implicit hooking, if only corfu implements such a formatter... 

> The kind-badge function can probably be used by Company too, even if it does not support affixation-functions.

I see.  I guess that could work.  Would certainly save the trouble of hacking into completion-in-region, since Corfu has the metadata goods like company-kind, after all.  Affixation function is sort of attractive because if other tools appear which support it, it might "just work".

jdtsmith:
> As for the name it may be better to have icons or badge in the name. Badge does not sound very familiar, so a better name would be kind-icons.

corfu also does not sound very familiar ;).  Sometimes that's an advantage.  I guess most people will want icons after all, so maybe `kind-icons` is OK.

minad:
> corfu also does not sound very familiar ;). Sometimes that's an advantage. I guess most people will want icons after all, so maybe kind-icons is OK.

I would go with corfu-kind if your package works only well with corfu. But for example if plugging into company works too you could go with kind-icons.

minad:
I have the in progress implementation with the kind formatter here: https://github.com/minad/corfu/tree/kind-formatter. I still have to think a bit more about it.

jdtsmith:
> here: https://github.com/minad/corfu/tree/kind-formatter

Interesting.  Maybe think thinspace here (also works for text prefixes):

```elisp
                (setf (cadr x) (concat p " "))))
```

minad:
> Maybe think thinspace here (also works for text prefixes):

No for affixations space is better. And for the kind formatter no padding is added. Also note that we cannot use thin space thanks to the character width alignment currently.

Also the margin bothers me a bit. If I color it with the prefix face, then the entire prefix icon will look misaligned since it will have a width of margin+2 chars.

jdtsmith:
> > Maybe think thinspace here (also works for text prefixes):
> 
> No for affixations space is better. And for the kind formatter no padding is added. Also note that we cannot use thin space thanks to the character width alignment currently.

Space is certainly _not_ better for me.  It's too wide, will be asymmetric, and if it's hard-coded in, I can't configure that anywhere!  In terms of lining things up, offsetting by a thin space is the SAME as offsetting by a thin margin (which is just a colored thinspace after all!).  So candidate alignment underneath the "prompt" will be pixel perfect.  Unless perhaps you are concerned with text-based prefixes being aligned with buffer text?  That to me isn't so important.  

You could argue a space between candidate and suffix makes more sense than a thin space, and there I might agree.  

> Also the margin bothers me a bit. If I color it with the prefix face, then the entire prefix icon will look misaligned since it will have a width of margin+2 chars. (EDIT: Probably I should just make the margins on both sides configurable independently. Then one could set it to zero on the left if your icons are used.)

Right.  Preserving symmetry is why you may want to try margin + 2 chars + margin, and just add another margin-width to your x offset (as I was doing in this PR).  You can see that effect in the very large screen shot above.

minad:
> Space is certainly not better for me.

Space is not better for icons, but space is better for affixations since it makes sense to treat the prefix and suffix of the affixation function symmetrically. This is what lead me to think about the corfu-kind-formatter-function, such that we can treat icons separately.

jdtsmith:
Note that `kind-icon` can _mix_ spaces and icons.  Since icons are 2x1 chars, that's fine.  In either case I'd prefer a flexible spacing between prefix and candidate (which can after all course flex up to one or more spaces wide!).

minad:
> Note that kind-icon can mix spaces and icons. Since icons are 2x1 chars, that's fine. In either case I'd prefer a flexible spacing between prefix and candidate (which can after all course flex up to one or more spaces wide!).

I prefer to not add any spacing from the side of Corfu, since by default the annotation/affixation functions add spaces as separators themselves. Corfu stays out of the way and it gives more freedom to the author of the annotation/affixation functions. See also #75.

jdtsmith:
The big issue: without pixel-level width calculations, only corfu itself knows how to correctly offset partial-width spaces (that it itself adds).  So annotation/affixation functions better not add those in themselves.  Unless there's a trick I'm missing?

minad:
Closed in favor of #75.

PR #76, opened 2021-11-10
Variable width alignment

minad:
Not pixels but floating point values in units of char width are used here.
This is a bit simpler and probably flexible enough for our purposes.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -408,8 +408,8 @@ A scroll bar is displayed from LO to LO+BAR."
          (x (or (car pos) 0))
          (y (or (cdr pos) 0)))
     (corfu--make-frame
-     (- x lm (* cw off)) y
-     (+ (* width cw) lm rm) (* (length lines) ch)
+     (- x lm (round (* cw off))) y
+     (+ (round (* width cw)) lm rm) (* (length lines) ch)
      (mapconcat (lambda (line)
                   (let ((str (concat lmargin line
                                      (if (and lo (<= lo row (+ lo bar))) sbar rmargin))))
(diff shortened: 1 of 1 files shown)


minad:
Test function, which was not possible before:

~~~ elisp
;; 1 char icon, 0.15 margin
(setq corfu-kind-formatter (lambda (_)
                             (propertize
                              (concat
                               (propertize " " 'display '(space :width 0.15))
                               "#"
                               (propertize " " 'display '(space :width 0.15)))
                              'corfu-width 1.3
                              'face '(:background "cyan"))))
~~~

minad:
Alternatively one could also search for `(space :width float)` display properties but then the situation would get quickly out of hand, since the `:width` specifications can be so complicated (different units, formulas, etc).

corfu.el
```diff
@@ -611,15 +611,20 @@ A scroll bar is displayed from LO to LO+BAR."
...
   (setq cands
         (cl-loop for c in cands collect
                  (cl-loop for s in c collect
                           (replace-regexp-in-string "[ \t]*\n[ \t]*" " " s))))
-  (let* ((cw (cl-loop for x in cands maximize (string-width (car x))))
-         (pw (cl-loop for x in cands maximize (string-width (cadr x))))
-         (sw (cl-loop for x in cands maximize (string-width (caddr x))))
+  (let* ((cw (cl-loop for x in cands maximize (corfu--string-width (car x))))
+         (pw (cl-loop for x in cands maximize (corfu--string-width (cadr x))))
+         (sw (cl-loop for x in cands maximize (corfu--string-width (caddr x))))
```
minad, line 627: I am overly general here - probably the 'corfu-width property should only be used for the prefix. The other strings won't have it.

jdtsmith:
Time traveler and mind reader too, eh?  Quite the skillset ;).  I'll take a look.  An advantage of "I'm always 3 chars wide" is good centering for 1 char string, 2 char string or icon.

minad:
> Time traveler and mind reader too, eh?

Just pretending to be a mind reader, easy as a time traveler :-P

> "I'm always 3 chars wide" is good centering for 1 char string, 2 char string or icon.

But seriously, this was just your idea since you wanted margin+icon+margin and the margin was reasonably close with 0.6. The only sacrifice we make is for 1 char text "icons" - a sacrifice which is justified in my opinion, since icons will be more popular.

So I close this variable width PR for now - keeping it simple, yay! We can still revisit this later if needed!

PR #83, opened 2021-11-17
Allow opting out of sorting prefix-matched candidates first

daanturo:


minad:
Thank you for the suggestion! This feature can already be tweaked in a single line, but arguably a bit less conveniently.

~~~ elisp
(advice-add #'corfu--move-prefix-candidates-to-front :override (lambda (_ x) x))
~~~

I am a bit hesitant to add configuration options to tweak such details, which are not often requested and which are probably not that interesting for many people. This holds in particular if the functionality can be easily disabled as in this case. There are more "magical" candidate movements taking place in Corfu - will we end up with a configuration option for each of those?

https://github.com/minad/corfu/blob/705d0ca1fd0891dcbc82ceb5d68e4da27957f0e8/corfu.el#L559-L568

Since you are using Doom, I guess you are familiar with my other packages and know that I am a bit opinionated on what to add and what to add not. Like Doom, opinionated but stubborn - hopefully. ;) So I would like to ask how long you used Corfu and if you gave it an earnest try for a longer while. There are two points I want to make: 1. Did you try Corfu for a longer time with the setting on and with the setting off, are you convinced that you want to disable this? 2. Often contributors come, propose some PR or feature request and then go back again to their familiar tools (often for various reasons, not because Corfu is bad, but because of familiarity and maybe because of some minor issues which are not yet sorted out in Corfu due to its younger age). This means I can resolve the issue by doing nothing and just waiting until more users come along requesting the same.

I argue that this sorting feature makes a lot of sense (up to the extent to not even make this configurable) since completion is usually triggered based on a prefix, by construction of the capf mechanism. However since we allow Orderless completion the first token is of course not necessarily the prefix - the sorting and movement of prefixed candidates to the beginning is the compromise I made.

On the other hand I acknowledge that requiring users to use advices is not such a nice style, in particular since the customize interface is pretty nice for exploring a package (Try `C-h v corfu M-x embark-export`). Furthermore the prefix sorting is something additional on top of the existing completion mechanism, which may be unexpected. So I may very well say yes to this proposal.

daanturo:
`corfu` hasn't replaced `company` for me yet until icons support was added recently.
Since `orderless` came, `company`'s prefix-only matching "suddenly" disappeared. It was a bit awkward at first but then got familiar to me overtime so I decided to like it, because when we only vaguely remember a part (especially in the middle) of the symbol's name, matching anywhere will help a ton.

`corfu--move-prefix-candidates-to-front` being an internal function made me hesitant to advise it because the fear of breakage due to parameter change in the future.

Personally I will happily use the advise approach if the function's parameters are stable.

minad:
> corfu hasn't replaced company for me yet until icons support was added recently.

Okay, so very recent. Icon support has been released basically yesterday. And the icon package is not yet available in a package archive. ;)

I suggest we close this for now and revisit the issue in a while. I am looking forward to more of your feedback in the future.

> ...because when we only vaguely remember a part (especially in the middle) of the symbol's name, matching anywhere will help a ton.

Yes, but this is not prevented by the sorting. Corfu only prioritizes the prefix matches, which seems like a reasonable assumption to me for `completion-at-point`.

> Personally I will happily use the advise approach if the function's parameters are stable.

Of course the function is internal, so I am not making guarantees. But when looking at the parameters - what could possibly change? So I expect this to stay as is. But you may still want to try Corfu in both configurations, with prefix sorting and without - just to see which one works really better for you. Also note that there are backends which only return candidates which already match the prefix, so the current prefix sorting leads to more uniformity. 

A while ago I had the prefix sorting removed from Corfu, since there were some changes in Orderless, which allowed the first word to compile to a ^regexp which would then be passed for prefix filtering to the completion table. Then I tried to use such a style dispatcher and removed the sorting functionality from Corfu. Obviously this had the effect that only prefix matches were shown, so I went back to not interpreting the first word as prefix, subsequently missed the prefix sorting again and added it back.

jdtsmith:
I don't mind prefix-sorting, but doesn't this overrule the in-built metadata sorter (if any), or, more seriously, the backend's sorting?  For example, fancy LSP servers use AI now to sort candidates with "what you probably want" at the top.

minad:
> I don't mind prefix-sorting, but doesn't this overrule the in-built metadata sorter (if any), or, more seriously, the backend's sorting? 

Yes, it does. But there are other minor overrides. For example exact matches are also ordered first.

> For example, fancy LSP servers use AI now to sort candidates with "what you probably want" at the top.

As I argued the primitive intelligence of ordering the prefix candidates first is reasonable. But if you disagree this can be easily overwritten with the advice.

We discussed in the other issue about prefix completion vs non-prefix completion. I'd argue that most systems have prefix completion in mind. The approach taken by Corfu by default gives us the best of both worlds - it orders prefixes first since users usually trigger completion based on prefixes and backends may even generate candidates based on prefixes (e.g. cape-dabbrev). But with orderless, the non-prefix candidates are still available, only deprioritized.

Issue #45, opened 2021-07-24
Suggestion: "soft" highlight first implicit candidate when none are selected


jdtsmith:
Here's company's take (no selection):

![image](https://user-images.githubusercontent.com/93749/126868532-dc45a2e9-c665-4677-ada9-90ccdcd286ab.png)

I.e. the 1st candidate is _always_ selected no matter what.  That doesn't work for corfu, because of the auto-commit-if-candidate-selected feature, but another less prominent selection face could be used to convey "if you hit `[Ret]`, you will have the first match inserted" (but never auto-committed!).

minad:
I tried this a few days ago and I didn't really like it since it adds zero value, only visual noise.

I also thought about another UI mode: Do not show the current selection in the input line and highlight the first candidate directly. But I think I like the current way more. Supporting different frontend display modes is something I would like to avoid for Corfu. I generally want Corfu to be configurable where it is cheap to have (corfu-quit-..., corfu-commit-predicate, etc), but I would like to avoid adding intrusive configuration options which affect many places and introduce many code paths.

jdtsmith:
Perhaps I wasn't clear enough: what I envision would be a pure UI flourish solely within `corfu--popup-show` — if none are selected, highlight the first candidate faintly.  See #50.  

Here's how it looks (dark):

![image](https://user-images.githubusercontent.com/93749/128724442-f7c9d5c1-8df0-4104-9d83-aac9f0b1f4b9.png)

vs. actual selection:

![image](https://user-images.githubusercontent.com/93749/128724910-91202013-31f4-4975-85b1-8479fdf21ae3.png)

minad:
No, I really don't like this. We should not add this. Maybe a better idea comes up regarding the implicit preselection of the first candidate. I also experimented with showing an overlay which extends the candidate "find-a|lias-regexp", where the "lias-regexp" is shown in faint color. However the issue is that this works only for prefix completion.

jdtsmith:
Sure.  Up to you of course, but you might consider checking in with other users to see how much mental overhead having _nothing_ selected gives.  But maybe I've just been using company too long.

minad:
I argue it gives not much mental overhead - the popup already is the indication. I don't want to close the door on other alternatives now and when I tried it, I didn't like the noise of the additional highlighting. The issue is, the highlighting adds zero information, it is only eye candy and takes the focus from the prompt.

minad:
[...]

PR #88, opened 2021-11-22
Rework selection

minad:


jdtsmith:
This branch is not working super well here.  One thing I noticed right away was "aggressive auto-committing".  For example, in the main branch I can starting typing (in elisp-mode) `with-l` and see various options, but continuing to type `with-local-` simply narrows them.  In this branch, the same sequence auto-commits quite aggressively upon adding `oc` (_and_ adds the `c` afterwards).   It also seems orderless isn't working.  E.g. I can't use `with-l w` and refine down.  Instead it auto-commits to `with-local-quitw`.  

Maybe I'm missing something?

minad:
I assume you did not install it correctly. Try emacs -Q without any of your customizations.

minad:
@jdtsmith Did you try again? This branch works quite well for me and should address all the issues discussed in #45.

jdtsmith:
I compiled it directly in the buffer, so not an installation issue.  I think rather that it does not cooperate anymore with my commit predicate:

```elisp
  (defun my/corfu-commit-predicate ()
    "Auto-commit candidates:
1. If a selection was made, aside from entering SPACE.
2. Just one candidate exists, and we continue to non symbol info. 
3. The 1st match is exact, and we have continued typing but
   haven't waited longer than `corfu-quit-no-match`"
    (cond
     ((seq-contains-p (this-command-keys-vector) ?.))

     ((>= corfu--index 0) ; a selection was made
      (not (seq-contains-p (this-command-keys-vector) ? )))

     ((eq corfu--total 1) ;just one candidate
      (seq-intersection (this-command-keys-vector) [?: ?, ?\) ?\] ?\( ? ]))
     
     ((and corfu--input ; exact 1st match
	   (string-equal (substring (car corfu--input) corfu--base)
			 (car corfu--candidates)))
           ;; (or (not (numberp corfu-quit-no-match))
	   ;;     (< (- (float-time) corfu--auto-start) corfu-quit-no-match))
      (seq-intersection (this-command-keys-vector) [?: ?. ?, ?\) ?\] ?\" ?' ? ]))))
```

Removing that, it behaved more normally, but I don't want to lose this behavior.

minad:
> Removing that, it behaved more normally, but I don't want to lose this behavior.

You won't lose this behavior, but you obviously have to rework your predicate. Please let's not discuss your user customizations here. I am interested in feedback regarding the reworked default behavior.

jdtsmith:
I think I should remove the "selection was made" test...

jdtsmith:
It's very relevant if you break customizations that are advertised in the Wiki.  I'll test for a while without a commit predicate and let you know.

minad:
> I think I should remove the "selection was made" test...

Yes, or you should rather replace it with `(/= corfu--index corfu--preselect)`.

> It's very relevant if you break customizations that are advertised in the Wiki. I'll test for a while without a commit predicate and let you know.

It is not documented in the wiki. But please go forward and add it there.

minad:
But I should make this clear - nothing which is documented in the wiki is in any way supported by the project itself. These are user-contributed additions. Breaking them can happen at every time. This also applies to my other packages.

jdtsmith:
Of course, but if certain customizations are in common use, it's a courtesy to see if they can avoid being broken!  [Update: glad to see you agree]. I was convinced we had documented some commit predicates there.  Perhaps it was in an issue.  

In any case, with that correction it's indeed working much better.  The one small oddity I noticed is that you can never actually "select" the first candidate.  You go right back to the buffer text if you arrow back up to the first candidate.   Not sure if you could stop first on the actual candidate, then move to the buffer text when it falls "off the top".   E.g. if I "really want the first candidate" and want to continue on typing say "." with auto-commit, in the old setup, I could arrow down, then press ".".  Now I have to hit return and do the same, which makes it distinct from all other candidates...

But overall this does represent an improvement and will be better I think for users coming from company/etc., since it's clear that the first candidate is special/implicitly selected upon first popup.

BTW, since I've been advertising how to setup kind-icon using `corfu-background`, and it seems that that's disappeared; it would be nice to make an alias there.

minad:
> In any case, with that correction it's indeed working much better. The one small oddity I noticed is that you can never actually "select" the first candidate. You go right back to the buffer text if you arrow back up to the first candidate. Not sure if you could stop first on the actual candidate, then move to the buffer text when it falls "off the top". E.g. if I "really want the first candidate" and want to continue on typing say "." with auto-commit, in the old setup, I could arrow down, then press ".". Now I have to hit return and do the same, which makes it distinct from all other candidates...

Yes, this is true. Note that the number of keypresses stays the same. The inconsistency in the handling of the first candidate is also present in the released implementation, where we have the "implicit selection". There is no free lunch.

But if you want to make it consistent you can set `corfu-preview-current=nil` and `corfu-commit-predicate=nil`. The inconsistency is a result of the auto committing concept.

> But overall this does represent an improvement and will be better I think for users coming from company/etc., since it's clear that the first candidate is special/implicitly selected upon first popup.

This is what I want to know. There is another edge case where this branch works better, if you are in eshell and complete a path, the explicit selection makes it clear if the path or the prompt is being accepted. This matters if you complete `mkdir /path/...` vs `cd /path/...`. For `mkdir` the prompt is accepted, since a new, non-existing name is allowed. For `cd` the directory must exist.

> BTW, since I've been advertising how to setup kind-icon using corfu-background, and it seems that that's disappeared; it would be nice to make an alias there.

`corfu-background` is deprecated and has been replaced by `corfu-default`.

jdtsmith:
Playing with CAPE using this rework branch I just noticed that on-demand completion (`M-Tab`) does not pre-select the first candidate, or indeed allow "return to commit the first candidate".  Is that intentional, and because the in-buffer text is empty?  This is also a difference (but of course now it is clear by whether the selection is "there" or not).

I think I get your eshell point.  Just tried and found that it never pre-selects the first candidate (in mkdir), but does in cd.  Clever.  

> corfu-background is deprecated and has been replaced by corfu-default.

Yes I see that, but since it's included on my config example for kind-icon, I'm asking if you can make an alias for it.  I'll update the recommendation, but it will break for existing users (as it just broke for me).

jdtsmith:
OK playing with this issue a bit more, I do find it somewhat confusing that _sometimes_ I can navigate "off the top", back to the (blank) buffer text, with nothing at all selected, but _other times_, navigating back to the 1st candidate puts me back on the buffer text, with the 1st selected.  I suspect this will confuse users/seem like a bug.  Note that in file-completion you get quite a few "blank buffer" completions.

minad:
> Playing with CAPE using this rework branch I just noticed that on-demand completion (M-Tab) does not pre-select the first candidate, or indeed allow "return to commit the first candidate". Is that intentional, and because the in-buffer text is empty? This is also a difference (but of course now it is clear by whether the selection is "there" or not).

You should tell me the completion table you are using. It depends on the backend. If the input string itself is a valid completion (like for mkdir), the first candidate is not selected, otherwise it is in order to force selection.

> I think I get your eshell point. Just tried and found that it never pre-selects the first candidate (in mkdir), but does in cd. Clever.

Thank you for confirming. This behavior is already present in the current implementation in the main branch. But it is entirely implicit and unclear. This is also what convinced me to rework the preselection and to get rid of the implicit mechanism.

> Yes I see that, but since it's included on my config example for kind-icon, I'm asking if you can make an alias for it. I'll update the recommendation, but it will break for existing users (as it just broke for me).

I understand. It has been deprecated and I will remove it in 0.16. It is better to remove and update this sooner than later, since Corfu does not have widespread support in themes.

jdtsmith:
I see this behavior with `cape-file-capf` in a lisp-buffer.  Wish I'd caught that deprecation I'll update the README.

minad:
> I see this behavior with cape-file-capf in a lisp-buffer.

If I type "/et", then "/etc" is selected. If I type "/etc/", then the files in "/etc/" are shown and nothing is selected since the input "/etc/" is already valid. This is the expected behavior.

jdtsmith:
Here's what I mean:

/`M-Tab` : nothing selected, can navigate "on and off" the top
/e`M-Tab`: /etc selected, cannot navigate "off the top"

minad:
> /M-Tab : nothing selected, can navigate "on and off" the top /eM-Tab: /etc selected, cannot navigate "off the top"

Yes, this is expected. 

* "/" is a valid name, so you can navigate to it and away from it to select a child. 
* If you enter "/e" it is not a valid name and you cannot navigate to it. The first candidate "/etc" is selected.

minad:
If this is good behavior is debatable of course, but note that you can always press C-g to exit with the current input, e.g., "/e" or "object." in python.

jdtsmith:
OK, I think it's reasonable once you understand it.  Is this backend specific?  How does corfu know it's dealing with files?

minad:
> OK, I think it's reasonable once you understand it. Is this backend specific? How does corfu know it's dealing with files and whether or not they exist?

Corfu asks the backend if the current input string is valid via `test-completion`. If the current input string is valid, it is preselected (corfu--index=corfu--preselect=-1). Otherwise the first candidate is selected (corfu--index=corfu--preselect=0).

As an alternative I could always allow movement to the input prompt line. But I am not sure if this would lead to more inconsistency. One has to consider what to do with cycling (corfu-cycle=t), move to the prompt then? Furthermore as soon as you enter new input, the selection will jump back to the first candidate which will be odd. I think it is better to keep this as is. Corfu always guarantees that you can quit with the current input with C-g, otherwise I would find it quite unusable, in particular with corfu-auto=t.

jdtsmith:
In LSP mode for example if I enter in the buffer exactly the text of one of the candidates, it does not get pre-selected.  Same thing in elisp-mode.  Right now I get "jumping back to 1st candidate" when entering additional text already with orderless.   Perhaps "jumping back to the buffer itself" is the more sensible behavior, but presumably comes with some other costs.

jdtsmith:
Honestly, it's somewhat hard to know what to do with the prompt.  For traditional in-buffer completion, there is ONLY the prompt text, and it never gets disturbed until you actively commit to a new candidate (company does this).   I can see the appeal of the "live preview" that corfu does to (though some users will be turned off). 

Anyway, have to run but happy to try any other permutations.

minad:
> In LSP mode for example if I enter in the buffer exactly the text of one of the candidates, it does not get pre-selected. Same thing in elisp-mode. 

This is also expected and correct. If you enter the exact candidate it is a valid completion obviously and as such it is selected. But it may feel a bit odd, so I can special case this and avoid movement to the prompt.

jdtsmith:
I'm having trouble reconciling these two:

> If you enter the exact candidate it is a valid completion obviously and as such it is selected.

> If the current input string is valid, it is preselected (corfu--index=corfu--preselect=-1).

minad:
> I'm having trouble reconciling these two:

I try to reformulate: 

If you enter the exact candidate as input, the input is a valid completion. Then the input/prompt is preselected (corfu--preselect=-1). More generally, if the backend accepts the current input string as valid completion, the input/prompt is preselected (corfu--preselect=-1).

Only if the current input is not a valid completion, but for example an orderless filter string with spaces, then the first candidate is selected (corfu--preselect=0).

minad:
But note that I added the special casing. The rules are now the following:

1. If the input is equal to the first candidate, the first candidate is preselected (corfu--preselect=0).
2. If the backend accepts the current input string as a valid completion, the input/prompt is preselected (corfu--preselect=-1).
2. Otherwise the first candidate is preselected (corfu--preselect=0).

jdtsmith:
I see so the blank string is a "valid" for a file completer, but not equal to any candidate, so no candidate is pre-selected.  

The equivalent for file completion is /etc`[M-Tab]` (prompt selected), but because a directory has a "/" at the end, you can never achieve your rule # 1, so it appears to be inconsistent.    That's actually unfortunate, because you'd like to do /etc`[M-Tab]` then `Tab` to go ahead with this selection to /etc/{candidate}.  

BTW your cape-file capf is much better than the comint one, since it only tries _on actual filepaths_!  Thanks for those.

minad:
> The equivalent for file completion is /etc[M-Tab] (prompt selected), but because a directory has a "/" at the end, you can never achieve your rule # 1, so it appears to be inconsistent. That's actually unfortunate, because you'd like to do /etc[M-Tab] then Tab to go ahead with this selection to /etc/{candidate}.

Ouch, you are right. This needs yet another special case only for files this time :(

I already special case files, in both Corfu and Vertico:

https://github.com/minad/corfu/blob/8a1101e6a12cb8cc43722a5808de2e0c338f54a9/corfu.el#L574-L575

Generally I dislike adding any kind of special casing, but well file paths are kind of important. I have to bow to the realities. However I somehow believe that something is wrong with the completion machinery in this regard - something is wrong with these slashes as part of the candidate. They should not be there or this should be solved somehow differently on a deeper layer.

minad:
I added the more complicated preselection check in https://github.com/minad/corfu/pull/88/commits/fb23cf993e5571fede582ec490213c25e9bd0224. Thank you for taking such a close look!

PR #100, opened 2022-01-05
Bind [return] instead

jdtsmith:
Binding to the key-code 13 (`C-m`) takes lower precedence than the vector key `[return]`. If a mode map bind the latter, `corfu-input` is not invoked by hitting Return when corfu is popped up.  See http://xahlee.info/emacs/emacs/emacs_key_notation_return_vs_RET.html.  

Try:

``` emacs-lisp
(local-set-key [return] (lambda () (interactive) (message "Overridden!")))
```

in `*scratch*` then use corfu there.


minad:
It is bad style to bind return because of this precedence issue. The modes which bind return should be fixed instead.

Issue #103, opened 2022-01-11
Extension wishlist


minad:
- [x] corfu-history (Prototype in https://github.com/minad/corfu/tree/extensions)
- [x] corfu-info (Prototype in https://github.com/minad/corfu/tree/extensions, Move out `corfu-show-documentation/location` from corfu.el, make it persistent and live updating)
- [x] corfu-quick (#127)
- [x] corfu-indexed (#127)
- [x] corfu-doc by @galeo, see https://github.com/galeo/corfu-doc/issues/8. Merged as corfu-popupinfo.
- [ ] corfu-mouse (See https://codeberg.org/akib/emacs-corfu-mouse, waiting on @AkibAzmain's copyright assignment)
- [ ] corfu-narrow (Proposal in #147)

Luis-Henriquez-Perez:
I take a similar approach to `vertico-quick` and modify the `corfu--format-candidates` function. I add `quick-letters` keys to the prefix of each candidate and am calling `corfu--format-candidates` with the updated candidates.

```elisp
(defun corfu-quick--format-candidates (orig candidates)
  "Format candidate, see `corfu--format-candidate' for arguments."
  (let ((updated-candidates nil))
    (pcase-dolist (`(,candidate ,prefix ,suffix) candidates)
      (setq quick-letters (corfu-quick--letters))
      (push (list candidate (concat quick-letters " " prefix) suffix) updated-candidates))
    (cl-callf reverse updated-candidates)
    (funcall orig updated-candidates)))
```

However, I'm having some trouble updating the `corfu` popup to display the candidates prefixed by the quick keys. What's `corfu`'s counterpart to `vertico--exhibit`? I thought it would be `corfu--update`. However, even though the candidates format is correctly updated by `corfu-quick--format-candidates` the popup disappears after calling `corfu-update`.

```elisp
(defun corfu-quick--read ()
  "Read quick key given FIRST pressed key."
  (cl-letf* ((old-fn (symbol-function #'corfu--format-candidates))
	     (new-fn (apply-partially #'corfu-quick--format-candidates old-fn))
	     ((symbol-function #'corfu--format-candidates) new-fn))
    (message "RESETTING")
    (corfu--update)
    1))
```

minad:
> I take a similar approach to vertico-quick and modify the corfu--format-candidates function. I add quick-letters keys to the prefix of each candidate and am calling corfu--format-candidates with the updated candidates.

[...]

PR #110, opened 2022-01-25
add corfu-indexed

Luis-Henriquez-Perez:
I've commented that I had made significant progress on corfu-indexed in #103. The functionality is almost done. I'd like to formally review it. I plan to rebase these commits to one as well.


minad:
I don't have much time for the next few weeks. I will review it after that. But this looks good. Feel free to give corfu-quick a stab in the meantime.

Luis-Henriquez-Perez:
> I don't have much time for the next few weeks. I will review it after that.

No worries. It happens. Take your time.

> Feel free to give corfu-quick a stab in the meantime.

Sure thing! :)

corfu-indexed.el
```diff
@@ -0,0 +1,89 @@
...
+(require 'corfu)
+(require 'pcase)
+
+(defface corfu-indexed
+  '((t :height 0.75 :inherit font-lock-comment-face))
+  "Face used for the candidate index prefix."
+  :group 'corfu-faces)
+
+(defvar corfu-indexed--commands
+  '(corfu-insert))
```
minad, line 43: Add corfu-complete here?

corfu-indexed.el
```diff
@@ -0,0 +1,89 @@
...
+;; We need to know this information in `corfu-indexed--handle-prefix', but at
+;; that time we don't have access to the list of candidates, hence this
+;; variable.
+(defvar-local corfu-indexed--max 0)
+;; No corfu indexed min.
+
+(defun corfu-indexed--format-candidate (orig cands)
+  "Format candidate, see `corfu--format-candidate' for arguments."
```
minad, line 51: There is no corfu--format-candidate. I wonder if we should introduce one, which is called by corfu--format-candidates. In Vertico I have vertico--arrange-candidates and vertico--format-candidate. Maybe such a scheme makes also sense here? What do you think?

Luis-Henriquez-Perez: > In Vertico I have vertico--arrange-candidates and vertico--format-candidate. Maybe such a scheme makes also sense here? What do you think?

It might make it more convenient to only have to consider an individual candidate to be formatted than a group of them. But, honestly, I'm don't have a strong opinion on it--at least not right now.

corfu-indexed.el
Luis-Henriquez-Perez: Ok. done.

minad:
Any progress on corfu-quick? Ideally I'd like to merge these two together, such that we see if changes to corfu.el itself are necessary.

Luis-Henriquez-Perez:
> Any progress on corfu-quick? 

Not yet. I'll can get it done within two days.

Luis-Henriquez-Perez:
Oh I tried to push my changes adding corfu-quick and github automatically closed this issue. Perhaps because I did so on a different branch. I did not close this intentionally.

Luis-Henriquez-Perez:
I don't have the permissions to reopen this issue. So if you could re-open it that would be great. My apologies I had no idea issues could be automatically closed like this.

I pushed the changes from `corfu-quick` here because I thought you wanted them both in the same PR, but let me know if that is not the case. 

`corfu-quick` is almost done. A quick summary:

`corfu-quick--format-candidates` - works.

`corfu-quick--letters` - should work, but I need to test indices at end of string length. I'll finish testing today.

`corfu-quick--read` - `(corfu--candidates-popup (point))` stopped working for me. The popup wouldn't show. I need to try it again and investigate it by going to an earlier commit where it did work. Also maybe I need to deal with the case the user passes in out of range characters (or maybe `corfu-quick-jump` does this).

`corfu-quick-jump` - works.

I added `corfu-quick-exit` for being able to referesh the popup without the corfu-quick letters in case the user changes their mind. Initially, I wrote it for my own testing but it may be useful all on its own.

minad:
I cannot reopen. Please create a new PR.

Issue #119, opened 2022-02-05
New approach for quitting on field boundary


minad:
OK, I have a corfu-auto + orderless setup that works very well for me.  Happy to put it on the wiki if it isn't too esoteric.  The idea is:

- Enable quit-on-boundary.
- Use a special unicode word-syntax character (in addition to space) for the orderless component separator.  "[ \u268b]" is my current choice (that char looks like `⚋`).  
- Bind `M-SPC` in corfu-map to insert this special character.
- Advise `corfu--post-command` to turn _off_ (via let-bind) `corfu-quit-at-boundary` _if the input contains this special separator character_. 

This works really well in practice.  All I have to do is remember to `M-SPC` instead of `SPC` with corfu _the first time only_ when I want to use multiple orderless filters.  And the default quit-on-boundary basically eliminates _all_ the false-positives, e.g. defining new variables, random words in strings or comments that my cape-ified dabbrev-enabled capf throws, etc.  But still lets you use full orderless power with regexps, etc. in buffer.  The only small drawback is the potential for unicode chars to persist in your buffer (if you `C-g` with a multi-component filter), hence my interest in a quick undo-and-quit, above. 

Thoughts?  Suggesting an advice is obviously not ideal for the wiki. Is there a predicate config option you are considering which might work instead for this `corfu--post-command` `:around` advice?

```elisp
  (defun my/corfu-disable-quit-with-orderless-advice (func)
    (let ((corfu-quit-at-boundary
	   (not (seq-contains (car corfu--input)
			      my/orderless-unicode-separator))))
      (funcall func)))
```

_Originally posted by @jdtsmith in https://github.com/minad/corfu/issues/109#issuecomment-1030686690_

minad:
See discussion following in #109 https://github.com/minad/corfu/issues/109#issuecomment-1030688695

jdtsmith:
Here's another nice unicode option (`#x00b7`, MIDDLE DOT):

<img width="163" alt="image" src="https://user-images.githubusercontent.com/93749/152661198-26b097a8-f374-486c-886b-2e434753a13d.png">

minad:
Why not just use space? If we dynamically turn off the boundary check for M-SPC we can do that.

jdtsmith:
OK, yeah, better; so just check the command keys?  We of course still want a normally-typed space to quit.

minad:
[...]

PR #120, opened 2022-02-06
Use separator character for boundary quit logic and prefix sort

jdtsmith:
Implement a component separator character and associated insertion command (`corfu-insert-separator-char`), used to inhibit quitting at completion boundaries (see #119).  This character is also removed from prefix sorting, superseding #118.

This is useful for multi-component completion styles such as orderless.  A convenient key binding for `corfu-insert-separator-char` such as `M-SPC` can be used to insert the initial completion component separator.

Obsoletes `corfu-quit-at-boundary` (for v0.19).  Setup documentation is provided.

### DIFF
--- a/README.org
+++ b/README.org
@@ -242,6 +242,10 @@ Then, when a new orderless component is desired, simply use =M-SPC=
 input, and arbitrary orderless search terms and new separators can be entered
 thereafter.
 
+Note that ~corfu-separator~ replaced and obsoleted ~corfu-quit-at-boundary~. If you
+want to restore the old behavior of ~corfu-quit-at-boundary=t~ you can bind
+~corfu-insert-separator~ to =SPC= in ~corfu-map~.
+
  #+begin_src emacs-lisp
    (use-package corfu
      ;; Orderless customizations
@@ -294,6 +298,7 @@ moves to the next candidate and further input will then commit the selection.
   - =RET= -> ~corfu-insert~
   - =M-g= -> ~corfu-show-location~
   - =M-h= -> ~corfu-show-documentation~
+  - =M-SPC= -> ~corfu-insert-separator~
   - =C-g= -> ~corfu-quit~
   - ~keyboard-escape-quit~ -> ~corfu-reset~
 
--- a/README.org
+++ b/README.org
@@ -224,11 +224,23 @@ sane Eshell experience.*
 
 ** Orderless auto-completion
 
-[[https://github.com/oantolin/orderless][Orderless]] is an advanced completion style that supports multi-component search filters separated by a configurable character (space, by default).  Normally, entering characters like space which lie outside the completion region boundaries (words, typically) causes corfu to quit.  This behavior is very helpful with auto-completion, which may pop-up when not desired, e.g. on entering a new variable name.  Just keep typing and corfu will get out of the way.  
+[[https://github.com/oantolin/orderless][Orderless]] is an advanced completion style that supports multi-component search
+filters separated by a configurable character (space, by default). Normally,
+entering characters like space which lie outside the completion region
+boundaries (words, typically) causes corfu to quit. This behavior is very
+helpful with auto-completion, which may pop-up when not desired, e.g. on
+entering a new variable name. Just keep typing and corfu will get out of the
+way.
 
-But orderless search terms can contain any characters; they are regular expressions. To use orderless in the buffer with ~corfu-auto~, set ~corfu-separator-char~ (a space, by default) to the primary character of your orderless component separator.
+But orderless search terms can contain any characters; they are regular
+expressions. To use orderless in the buffer with ~corfu-auto~, set
+~corfu-separator-char~ (a space, by default) to the primary character of your
+orderless component separator.
 
-Then, when a new orderless component is desired, simply use =M-SPC= (~corfu-insert-separator-char~) to enter the /first/ component separator in the input, and arbitrary orderless search terms and new separators can be entered thereafter.  
+Then, when a new orderless component is desired, simply use =M-SPC=
+(~corfu-insert-separator-char~) to enter the /first/ component separator in the
+input, and arbitrary orderless search terms and new separators can be entered
+thereafter.
 
  #+begin_src emacs-lisp
    (use-package corfu
@@ -241,7 +253,7 @@ Then, when a new orderless component is desired, simply use =M-SPC= (~corfu-inse
      ;;	   ("M-SPC" . corfu-insert-separator-char))
      :init
      (corfu-global-mode))
- #+end_src    
+ #+end_src
 
 ** TAB-and-Go completion
 
--- a/README.org
+++ b/README.org
@@ -46,7 +46,7 @@
   - The selected candidate automatically committed on further input by default
     (configurable via ~corfu-commit-predicate~).
   - The [[https://github.com/oantolin/orderless][Orderless]] completion style is supported. The filter string can contain
-    arbitrary characters, including spaces, if ~corfu-quit-at-boundary~ is nil.
+    arbitrary characters, including spaces, if ~corfu-separator-char~ is configured to match the orderless separator, and a special key is used to insert the first separator.
   - Deferred completion style highlighting for performance.
   - Jumping to location/documentation of current candidate.
   - Show candidate documentation/signature string in the echo area.
@@ -77,13 +77,15 @@
       ;; (corfu-cycle t)                ;; Enable cycling for `corfu-next/previous'
       ;; (corfu-auto t)                 ;; Enable auto completion
       ;; (corfu-commit-predicate nil)   ;; Do not commit selected candidates on next input
-      ;; (corfu-quit-at-boundary t)     ;; Automatically quit at word boundary
+      ;; (corfu-separator-char ? )      ;; Inihibit quit at word boundary with this char
       ;; (corfu-quit-no-match t)        ;; Automatically quit if there is no match
       ;; (corfu-preview-current nil)    ;; Disable current candidate preview
       ;; (corfu-preselect-first nil)    ;; Disable candidate preselection
       ;; (corfu-echo-documentation nil) ;; Disable documentation in the echo area
       ;; (corfu-scroll-margin 5)        ;; Use scroll margin
 
+      ;; :bind (corfu-map               ;; Bind a command to insert 1st separator char
+      ;;        ("M-SPC" . corfu-insert-separator-char))
       ;; You may want to enable Corfu only for certain modes.
       ;; :hook ((prog-mode . corfu-mode)
       ;;        (shell-mode . corfu-mode)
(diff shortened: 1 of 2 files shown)


minad:
Very good, I take a look at this in the next few days. So if we bind SPC to corfu-insert-separator we get the current quit-at-boundary behavior? If we bind M-SPC instead we get the enhanced behavior? I guess we could bind M-SPC by default.

README.org
```diff
@@ -104,14 +106,14 @@
...
-            completion-category-overrides '((file (styles . (partial-completion))))))
+	    completion-category-defaults nil
+	    completion-category-overrides '((file (styles . (partial-completion))))))
 
     ;; Use dabbrev with Corfu!
     (use-package dabbrev
       ;; Swap M-/ and C-M-/
       :bind (("M-/" . dabbrev-completion)
-             ("C-M-/" . dabbrev-expand)))
+	     ("C-M-/" . dabbrev-expand)))
```
minad, line 116: Quick review. Please ensure that you don't accidentally change ws.

README.org
```diff
@@ -77,13 +77,15 @@
...
-      ;; (corfu-quit-at-boundary t)     ;; Automatically quit at word boundary
+      ;; (corfu-separator-char ? )      ;; Inihibit quit at word boundary with this char
       ;; (corfu-quit-no-match t)        ;; Automatically quit if there is no match
       ;; (corfu-preview-current nil)    ;; Disable current candidate preview
       ;; (corfu-preselect-first nil)    ;; Disable candidate preselection
       ;; (corfu-echo-documentation nil) ;; Disable documentation in the echo area
       ;; (corfu-scroll-margin 5)        ;; Use scroll margin
 
+      ;; :bind (corfu-map               ;; Bind a command to insert 1st separator char
+      ;;        ("M-SPC" . corfu-insert-separator-char))
```
minad, line 88: Bind M-SPC by default.

corfu.el
```diff
@@ -83,11 +83,20 @@ The value should lie between 0 and corfu-count/2."
...
+(defcustom corfu-separator-char ? 
+  "Component separator character.
+The character used for separating components in the input.
+Useful for multi-component completion styles such as orderless.
+To use, bind `corfu-insert-separator-char' to a convenient
+key (e.g. M-SPC) in corfu-map, and use this binding to enter the
+first separator character.  Both this command and the presence of
+this separator character within the input will inhibit quitting
+on completion boundaries."
+    :type 'character)
```
minad, line 99: Support character or nil here.

corfu.el
```diff
@@ -83,11 +83,20 @@ The value should lie between 0 and corfu-count/2."
...
-(defcustom corfu-quit-at-boundary nil
-  "Automatically quit at completion field/word boundary.
-If automatic quitting is disabled, Orderless filter strings with spaces
-are allowed."
-  :type 'boolean)
+(make-obsolete 'corfu-quit-at-boundary
+               "see the new `corfu-separator-char' customization."
+               "0.19")
+
+(defcustom corfu-separator-char ?
```
minad, line 90: Write ?\s

corfu.el
```diff
@@ -549,7 +558,8 @@ A scroll bar is displayed from LO to LO+BAR."
 
 (defun corfu--move-prefix-candidates-to-front (field candidates)
   "Move CANDIDATES which match prefix of FIELD to the beginning."
-  (let* ((word (replace-regexp-in-string " .*" "" field))
+  (let* ((word (replace-regexp-in-string
+		(format "[%c].*" corfu-separator-char) "" field))
```
minad, line 562: Here we should probably use a different method for string splitting at this character. (Or quote the character. But I would prefer a different splitting method, first find char and then take the substrings.)

minad: Only split when the separator-char is non-nil.

corfu.el
```diff
@@ -835,6 +845,13 @@ there hasn't been any input, then quit."
   "Return t if a candidate is selected and previewed."
   (and corfu-preview-current (/= corfu--index corfu--preselect)))
 
+
+(defun corfu-insert-separator-char ()
+  "Insert a separator character and inhibit quit on completion boundary.
+Bind to a convenient key in corfu-map."
+  (interactive)
+  (insert corfu-separator-char))
```
minad, line 853: Here we should probably throw an error if the separator char is nil?

corfu.el
```diff
@@ -848,8 +865,9 @@ there hasn't been any input, then quit."
                       (save-excursion
                         (goto-char beg)
                         (<= (line-beginning-position) pt (line-end-position)))
-                      (or (not corfu-quit-at-boundary)
-                          (funcall completion-in-region-mode--predicate))))
+                      (or (eq this-command 'corfu-insert-separator-char)
+			  (seq-contains-p (car corfu--input) corfu-separator-char)
+			  (funcall completion-in-region-mode--predicate))))
```
minad, line 870: Please fix the indentation, space only. We should also check if corfu-separator-char is nil.

corfu.el
minad: Hmm, I have a different idea. In order to effectively disable the separator character we can use ?\0. This keeps eveything simpler.

minad:
I gave it a quick review. This looks very good, nice and simple. There are only a few minor details. Then there is the question if we want to support disabling the separator character completely - by setting it to ?\0 or nil.

corfu.el
minad: Here I would prefer if you check with `search-forward` for `corfu-separator-char` instead of checking both this-command and `corfu--input`. This decouples the implementation slighly more, since we don't refer to `corfu-insert-separator-char`.

jdtsmith:
> So if we bind SPC to corfu-insert-separator we get the current quit-at-boundary behavior?

I suppose that's right.  Or, if you (or the user) binds nothing at all to `corfu-insert-separator-char`, you will simply fall through to the predicate boundary check.  So the default could become the less intrusive option.  

> if we want to support disabling the separator character completely

I tried to think of a real usage case for disabling the behavior of the separator.  Unless you invoke `M-SPC`, the separator is benign.  The only thing I could come up with:

1. You do not use orderless or a multi-component completion style. 
2. Your mode's CAPF happily includes the separator char within the completion boundary (I suppose more likely if it's not space).
3. You want _other_ non-word chars to quit completion (e.q. punctuation like `,` or `=`).

I don't know if 2 is occurring anywhere in the wild for the default space separator.  But allowing this to be `nil` is perhaps easy insurance for any cases we are missing.  I'd still recommend keeping space as the default though.

Also, regarding the binding, should we bind it ourselves or let the user pick?  I've also tried `S-SPACE` which works well.

corfu.el
jdtsmith: One related issue that occurred to me: `orderless-component-separator` is not a char but a string, which is passed as-is to `split-string`. It's interpreted as a regexp.  We _could_ follow that approach.  Then use `split-string` ourselves for prefix sorting _and_ detecting the separators.

corfu.el
jdtsmith: Yes, but see below for char-vs-string.

corfu.el
jdtsmith: That would be simple since it should never match in buffer.

corfu.el
jdtsmith: Update: but you'd still need the char to insert on `M-SPC`...

corfu.el
jdtsmith: I'm not sure I see how that would work.  We want M-SPC and SPC to do two _different_ things (continue, or quit), despite inserting exactly the same text in buffer.

minad: Right. I think I don't understand what your condition means since you check for the character being present in the input.

Originally I thought about introducing a variable (corfu--separator-present) which is set by corfu-insert-separator to allow a distinction. But maybe this is unnecessary?

minad:
I just want the ability to disable this out of principle, but I agree that having a separator like space (also having it enabled by default) is benign

corfu.el
minad: Hmm. Good point. But it seems that a single separator char is sufficient for most scenarios and most efficient. In comparison, I use escaped space where I mean space literally.

Let's go with a single character for now. We can still generalize later.

corfu.el
minad: So you say as soon as a separator is present we stay alive. Otherwise stay alive if the separator was just inserted by M-SPC or if the predicate is true.

This does not seem strictly correct to me. You could have the separator but you would still want the predicate to be checked. For example if I hit space twice I want the predicate checked for both spaces.

minad: Hmm. This needs more thought. I am not sure about my conclusion that orderless would interfere, since this is only about the capf boundary.

jdtsmith: The idea here is that the first M-SPC "gets you in" to no-boundary-check mode, and allows a space to be included at the end of the input.  corfu-update implicitly sets the new input _after_ these checks, so the this-command check is essential.  Once in the special mode, the presence of the separator character alone is sufficient to keep you in this mode. Then you can just type any orderless characters, including (regular) space.  If you delete all spaces, you lose this protection, and will have to M-SPC again.

The thinking is, if you went to the trouble of typing a special key command, you aren't likely to want to drop out of that special mode with another such separator.  Hence the mention of only needing to remember M-SPC "the first time".  

We could instead use an extra variable (like corfu--check-boundary), have it always default to on, and use a custom binding to toggle it.  But you'd have to remember to hit this before reaching for the space bar.  In practice, the existing solution seems to work really well.

jdtsmith: One other minor addendum: if your separator character is _allowed_ within your CAPF boundaries, you are good to go without doing anything special.  But that would be a strange setup indeed.

minad: Okay let's keep your approach for now. Maybe add a short comment? I like that we don't need another variable and it is nice that by deletion we go back to the unseparated mode.

corfu.el
jdtsmith: Using a new prefix separator method.

corfu.el
jdtsmith: Done (with `nil`).

corfu.el
jdtsmith: Went with `nil`; done.

corfu.el
```diff
@@ -558,8 +558,8 @@ A scroll bar is displayed from LO to LO+BAR."
 
 (defun corfu--move-prefix-candidates-to-front (field candidates)
   "Move CANDIDATES which match prefix of FIELD to the beginning."
-  (let* ((word (replace-regexp-in-string
-                (format "[%c].*" corfu-separator-char) "" field))
+  (let* ((word (substring field 0
+                          (seq-position field corfu-separator-char)))
```
minad, line 563: Add a small optimization here please. If the character hasn't been found, do nothing. We can improve this code while we are on it :)

minad:
> Also, regarding the binding, should we bind it ourselves or let the user pick? I've also tried S-SPACE which works well.

We should add the binding M-SPC by default. It is easy to unbind this in a user configuration. We also preconfigure separator=SPC.

README.org
jdtsmith: done

corfu.el
jdtsmith: `nil` supported.

corfu.el
jdtsmith: done

corfu.el
jdtsmith: simpler method used.

corfu.el
jdtsmith: done.

corfu.el
jdtsmith: Are you sure that small optimization doesn't already happen inside of `(substring field 0 nil)`?

minad: Disregard, I got confused.

minad:
This looks good, thanks!

corfu.el
jdtsmith: I checked, and a small optimization does actually save 10-20%. Done.

PR #122, opened 2022-02-07
corfu-quit-no-match: Remove support for seconds

minad:
This PR removes the timer support from `corfu-quit-no-match`. I wanted to ping you first about this since you proposed the timer based solution. I think we are converging to a better solution now with `corfu-separator`, so maybe it is justified to remove this?

One idea I had is to support three different values instead of the timer:

- corfu-quit-no-match = nil: Always stay alive.
- corfu-quit-no-match = separator: Only stay alive if there is no match and a separator has been inserted. At this point we are in "orderless filtering mode".
- corfu-quit-no-match = t: Always quit if there is no match.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -99,11 +99,9 @@ To enter the first separator character, call
 Useful for multi-component completion styles such as orderless."
   :type '(choice (const nil) 'character))
 
-(defcustom corfu-quit-no-match 1.0
-  "Automatically quit if no matching candidate is found.
-If a floating point number, quit on no match only if the auto-started
-completion began less than that number of seconds ago."
-  :type '(choice boolean float))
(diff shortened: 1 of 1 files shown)


jdtsmith:
I agree with this.  Lots of this earlier complexity was to support auto+boundary-quit=nil, and attempt to mitigate the resulting nuisance false-positives and lingering “No match” pop ups.  The new “opt-in” to multi-component solves most of this.  

But `‘separator` is a great idea.  While composing a regexp, or planning to put a `!` at the end of a component, you will often have no match, until you have them again.

minad:
Okay, thanks for the feedback! I also made `corfu-quit-no-match=separator` the default.

PR #124, opened 2022-02-07
Fix doc for restoring default behavior

jdtsmith:
The default behavior was quit-at-boundary=nil, and that's achieved by always "blocking" the boundary check, by binding your separator character to the new insert command.  This is only approximate, because other boundary-breakers like punctuation will still cause a quit in the 1st component.  But you won't need M-SPC to avoid quits (and will have lots of "false-positives").  The analogous setting for quit-at-boundary=t (_always_ quit at boundary) is corfu-separator=nil.

### DIFF
--- a/README.org
+++ b/README.org
@@ -241,9 +241,12 @@ Then, when a new orderless component is desired, simply use =M-SPC=
 input, and arbitrary orderless search terms and new separators can be entered
 thereafter.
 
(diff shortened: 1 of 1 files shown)


minad:
Oh thanks, this is even better! It is good that we added the ability to disable the separator entirely.

PR #132, opened 2022-02-21
Delay docsig function call with timer

### LINKED ISSUE #131: Initial and secondary delay for `corfu-echo-documentation`
I noticed two small issues with `corfu-echo-documentation`:

1. Because of the logic `(or (eq corfu-echo-documentation t) corfu--echo-message)`, once a single candidate has had its documentation shown (usually the first), all further candidates have their docs shown immediately, without a time delay.  
2. The `:company-docsig` function may be expensive, e.g. contacting the server/process to learn the documentation.  In the present design, this function is called on every `corfu--update`.

Ideally, the `corfu-echo-documentation` timer would apply to all candidates, and the function would be called "just in time" to avoid unnecessary server contact.   That way you could scroll rapidly through a list of candidates without concern, pausing to see the docs for one, if interested.

jdtsmith:
This implements a timer for calling the company-docsig function and displaying its results.  The impetus is to work more smoothly with costly docsig function which interact with servers or processes.    This uses only a single short delay, which I find works well, allowing you to scroll through many candidates without querying for docsig, then pausing briefly on any candidate to query and echo its short documentation.  Fixes #131.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -131,7 +131,7 @@ separator: Only stay alive if there is no match and
   "Width of the bar in units of the character width."
   :type 'float)
 
-(defcustom corfu-echo-documentation 0.2
+(defcustom corfu-echo-documentation '(1.0 0.2)
   "Show documentation string in the echo area after that number of seconds.
 Set to nil to disable docsig, or t to echo immediately on
 selecting a new candidate.  Can also be a 2-element list of
--- a/corfu.el
+++ b/corfu.el
@@ -134,8 +134,18 @@ separator: Only stay alive if there is no match and
 (defcustom corfu-echo-documentation 0.2
   "Show documentation string in the echo area after that number of seconds.
 Set to nil to disable docsig, or t to echo immediately on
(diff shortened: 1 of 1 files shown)


minad:
This doesn't work well for me. While I like a secondary short delay of 0.2s I wouldn't want an initial delay of only 0.2s. Therefore two delays.

The question is what to do if there is a visible docsig and the candidate changed. Immediately hide the message? Otherwise you will see an outdated docsig in the echo area.

jdtsmith:
I immediately hide message on new candidate, so you are never out of date.  For very short delays (~0.1s) this can lead to flashing as you arrow through candidates. 

I have implemented an initial/subsequent delay, either of which can be `t` (for instant).  Sounds like you may like `(1.0 t)` or similar.  I added a state variable `corfu--initial` to keep track of whether this is the first update.

minad:
Thanks, I simplified it a little bit in 7f5dacb340f0ec3f069744a76b3021b055ac98ed.

PR #137, opened 2022-02-28
.gitignore: Ignore corfu.texi

DamienCassou:
This file is automatically generated from README.org and must not be
commited.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/corfu.texi


minad:
I don't need a gitignore for now and most of my Emacs projects don't have one. Maybe it makes sense to add one? I am not convinced to be honest. In any case then it should include more than this single entry and it should be done consistently across all my Emacs projects.

PR #142, opened 2022-03-04
Allow enabling/disabling auto completion per-buffer

mattbeshara:
I find that Corfu's auto completion is very useful when editing code, but not so much when editing prose.  It seems that currently auto completion is an 'all or nothing' type deal, but I'd really like it if it could be only be enabled for specific buffers.  With the changes in this PR, it's possible to write a trivial `prog-mode-hook` function that enables auto completion, but leaves it disabled for other kinds of buffers.  It seems to work well, but I suppose if this PR or some other code like it was to be merged, it would also necessitate some changes to the documentation.  Would a change like this be considered for merging into Corfu, or is this customisation not likely to be found beneficial enough to others to warrant this change in behaviour?

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -177,7 +177,8 @@ The completion backend can override this with
 
 (defcustom corfu-auto nil
   "Enable auto completion."
-  :type 'boolean)
+  :type 'boolean
+  :local t)
 
 (defgroup corfu-faces nil
   "Faces used by Corfu."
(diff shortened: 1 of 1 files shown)


minad:
It is not necessary to change anything in Corfu. Disabling auto completion is already possible by setting `corfu-auto=nil` locally. It may be good to document this a bit better.

~~~ elisp
(add-hook 'text-mode-hook #'corfu-no-auto)
(defun corfu-no-auto ()
  (setq-local corfu-auto nil))
~~~

mattbeshara:
Alright, I'll try that.  It seemed that setting `corfu-auto=nil` didn't take effect when I tried that previously, but I think if I turn on Corfu in a hook rather than use `corfu-global-mode` that might address it.  Thanks!

PR #153, opened 2022-04-10
Add link to alternatives for more context

jgarte:


README.org
```diff
@@ -465,7 +465,7 @@ The command ~corfu-move-to-minibuffer~ is defined here in terms of
 
   - Corfu uses child frames to show the popup. For now Corfu falls back to the
     default setting of the ~completion-in-region-function~ on non-graphical
-    displays. You can use one of the alternatives in terminals.
+    displays. You can use one of the [[https://github.com/minad/corfu#alternatives][alternatives]] in terminals.
```
minad, line 468: This should be an org file internal link, not a link to gh. But I am not sure if we should add this link only. Are there other places where we would link across sections.

PR #160, opened 2022-04-26
Prevent auto-popup when first candidate is exact match.

jdtsmith:
One of the last rough edges of corfu with auto is it offers completions on already complete matches.  I.e. if I quickly type a full and complete symbol, e.g. `defun`, and pause at the end of that, corfu will pop up, offering, as its very first candidate, `defun`!  (Along with some others that end in `defun`...) 

This means I have to keep moving on quickly, or keep my `C-g` ready.  This PR adds a flag and slightly tweaks the logic when setup is called from `corfu--auto`, pre-quitting if the first candidate is an exact match (even if there are more candidates).  Then I can always `M-Tab` if I want to see all of those.  This does create a slight asymmetry between auto completion and command-requested completion, which I believe is well worth it in this case of mostly-false-positives.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -818,8 +818,9 @@ there hasn't been any input, then quit."
         (corfu--echo-show))
     (corfu--echo-show)))
 
-(defun corfu--update ()
-  "Refresh Corfu UI."
+(defun corfu--update (&optional auto)
+  "Refresh Corfu UI.
+AUTO is non-nil when called by auto completion."
   (pcase-let* ((`(,beg ,end ,table ,pred) completion-in-region--data)
                (pt (- (point) beg))
                (str (buffer-substring-no-properties beg end))
@@ -840,11 +841,16 @@ there hasn't been any input, then quit."
      ;; 1) Initializing, no candidates => Quit. Happens during auto completion.
      ((and initializing (not corfu--candidates))
       (corfu-quit))
-     ;; 2) Single exactly matching candidate and no further completion is possible.
+     
+     ;; 2) Exactly matching candidate during auto initialization,
+     ;; or single candidate with no further completion possible.
      ((and (not (equal str ""))
-           (equal (car corfu--candidates) str) (not (cdr corfu--candidates))
-           (not (consp (completion-try-completion str table pred pt corfu--metadata)))
-           (or initializing corfu-on-exact-match))
+           (equal (car corfu--candidates) str)
+	   (or (and initializing auto)
+	       (and (not (cdr corfu--candidates))
+		    (not (consp (completion-try-completion
+				 str table pred pt corfu--metadata)))
+		    (or initializing corfu-on-exact-match))))
       ;; Quit directly when initializing. This happens during auto completion.
       (if (or initializing (eq corfu-on-exact-match 'quit))
           (corfu-quit)
@@ -1144,7 +1150,7 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
                      table
                      (plist-get plist :predicate)))
          (corfu--setup)
-         (corfu--update))))))
+         (corfu--update t))))))
 
 (defun corfu--auto-post-command ()
   "Post command hook which initiates auto completion."


minad:
I don't really agree. Why do you pause? You can press space or something else and the candidate will not appear. Or just don't use orderless. You could also combine orderless with basic, e.g., (basic orderless) to prevent this from happening. But be aware of https://github.com/minad/consult/issues/559. Maybe we have to offer some means to easily overwrite the completion style for Corfu only.

jdtsmith:
> I don't really agree with the asymmetry. 

I should have used the term "hybrid":

1. `'always`: Always auto-complete with the full list of candidates.
2. `'hybrid`: Auto-complete only if the in-buffer string isn't itself a complete match.
3. nil: Never auto-complete, completion is on-demand with a keypress.

> Why do you pause?

Usually to think.  But seriously, this happens so often my brain is trained to be on alert for the likely false positive coming my way.  I used to use the commit predicate to stamp these out, but that functionality was removed.  

> Or just don't use orderless. 

In-buffer orderless is for me the most important feature of corfu.  So reverting to basic is a non-starter.  

But this gives me an idea: I could approximate the `'hybrid` style using orderless dispatch, i.e. "use basic on first term only if in an auto-popup".  For that I'd need a reliable means of knowing it's a corfu--auto. 

Would you consider the addition of a special `:corfu-auto t` to `completion-extra-properties` when completion originates from `corfu--auto-complete`?  I could then document how to do this via orderless config for people who struggle with this remaining form of false positive.

jdtsmith:
All fair points.  I suspect I'm in pretty rare company in terms of how I use corfu.  Pretty much all of the (far fewer than before!) false positives come from attempts to combine:

1. corfu-auto
2. orderless in buffer

Do you have a sense of how many corfu users do this?  As I understand it, you don't even use that combination (leaving auto disabled), right?  To me it's an easy to explain super-power of corfu that you can make this combination work reasonably well, and a clear distinction from company.  The new M-SPC really goes a long way there.  So it's possible the rough edges I find are particular to my workflow, but it's also possible that I'm one of the few working that way (and of course it's possible that situation will persist!).  

BTW, without corfu auto, this particular issue would never occur: I'd never _intentionally_ invoke completion after `(defun` unless I really wanted something _else_.  So that's a fundamental builtin asymmetry in auto- vs. on-demand completion, although not all may agree.  

I just tried company with substring (having never done so before), and it does not have any special workarounds, in fact doesn't sort `defun` to the front either.  

image

So corfu is no worse (even slightly better due to better sorting).  But I suspect most people use company only with basic (and maybe corfu too?).  In that case, company, like corfu, will not prompt for the single full match, and all is simple and easy.  Most company backends do not even respect completion styles as I understand it.

jdtsmith:
I'll draw one other distinction: popping up with a different set of candidates whether invoked from auto or on-demand is a true asymmetry, and _would_ be confusing.  _Not_ popping up automatically each time the timer ticks,  if it's unlikely the user wants that auto-pop to happen, is a different category.  You could imagine `corfu-auto=nil, t, conservative` where the latter only pops up if corfu-auto has something really important to show you.

minad:
I use corfu-auto=t with Orderless, but I use mostly the Corfu default settings.

minad:
> Not popping up automatically each time the timer ticks, if it's unlikely the user wants that auto-pop to happen, is a different category. You could imagine corfu-auto=nil, t, conservative where the latter only pops up if corfu-auto has something really important to show you.

I am not sure I agree. This is hair splitting. Not popping up means that there are no candidates available or no candidates except the one single candidate. Therefore not popping up is just the same as the asymmetry. Overall I think I cannot agree to your proposed change in behavior. 

From my experience, the issue you have here also doesn't bother me too much. As long as the exactly matching candidate appears first in the popup, no bad side effects are to be expected. The RET key is clobbered, so if I would want to enter a newline I would have to press RET twice which can be a minor annoyance. But that's all?

There is another scenario which is much more problematic. For example I have `cape-dabbrev` activated globally and when typing in text buffers (e.g. commit messages) or Org buffers it sometimes happens that I type some word (which doesn't appear as an exact match), hesitate for a while and then I press RET and it auto completes to a word which I didn't intent to type. This reminds me of the ugly behavior which is prevalent in mobile phones which always stupidly auto correct my "typos". This is something which truly bothers me and which should be fixed. I think the issue is that `cape-dabbrev` filters out candidates which are not sufficiently longer than the current input. Exact matches are filtered out, resulting in this bad side effect. I should probably change `cape-dabbrev` such that it still filters out candidates which are too short but takes care to preserve exact matches.

jdtsmith:
Thanks for entertaining this line of thought.

> Not popping up means that there are no candidates available or no candidates except the one single candidate.

Or that the timer didn't fire in time? I agree that if you explicitly _ask_ for candidates, not showing any certainly means this.  

> no bad side effects are to be expected

Just the cognitive overhead of having to dismiss an unwanted pop-up, and the feeling that corfu should clearly _know_ such a pop-up is unwanted.  But obviously that's a personal assessment.  Given that company doesn't even do this, I can certainly see your POV.  I can hack around it with some advice I suspect. 

> hesitate for a while and then I press RET and it auto completes to a word which I didn't intent to type.

Yes!  Bad autocorrect is a great analogy. Related genre to unwanted pop in my mind — worse, but rarer.  I also add in dabbrev candidates, which is mostly great, but notice the same "race" between corfu-auto and Ret, with unintentional insertions in comment strings, etc.  Perhaps a configuration setting to defeat the "race" like `corfu-return-means-quit-before`, set to 0.1s or so, with the stopwatch started as soon as the timer calls auto-complete.  

> One targeted solution is cape-capf-accept-all 

How does overriding `test-completion` to return t help the dabbrev issue when there is more than one dabbrev candidate?  

Another bother is dabbrev gives versions with and without a final period for me.  

image

minad:
> Or that the timer didn't fire in time? I agree that if you explicitly ask for candidates, not showing any certainly means this.

No, this is still all deterministic and predictable. `defun TAB` means the same as `defun WAIT`.

> Just the cognitive overhead of having to dismiss an unwanted pop-up, and the feeling that corfu should clearly know such a pop-up is unwanted.

This is where I disagree. I think it is clearly wanted. If not, use a different completion style. It is as simple as that.

> Related genre to unwanted pop in my mind — worse, but rarer. I also add in dabbrev candidates, which is mostly great, but notice the same "race" between corfu-auto and Ret, with unintentional insertions in comment strings, etc.

This race only ever occurs in a context where you want to press RET to insert a newline. From my experience this happens mostly in text and not so much in programming modes since there one often inserts operators.

> How does overriding test-completion to return t help the dabbrev issue when there is more than one dabbrev candidate?

It will help since the first candidate is never preselected. test-completion determines if the input/prompt itself is an acceptable candidate.

> Another bother is dabbrev gives versions with and without a final period for me.

Yes, this bothers me too. Dabbrev is really not good. But this is the fault of the Emacs implementation not of the Cape wrapper. I think in the end I will replace the Dabbrev Capf with a more efficient and less annoying custom implementation, but it is not the most interesting thing to hack on :-P

Issue #149, opened 2022-03-29
Popup size is wrong if the height of the `corfu-default` face is modified 


minad:
Example 1:
~~~
(set-face-attribute 'corfu-default nil :height 0.5)
~~~

Example 2:
~~~
(set-face-attribute 'corfu-default nil :height 100)
~~~

This issue also occurs if different font families are mixed. See #148 for the original report.

rakyi:
Not sure if it has the same root cause, but the popup size is also wrong with e.g. `(setq default-text-properties '(line-spacing 0.15 line-height 1.15))`, which is useful to change the line height of most lines (taken from https://github.com/syl20bnr/spacemacs/issues/10502#issuecomment-404453194).

zhenhua-wang:
It is possible to compute the popup size based only on fixed-pitch face? I tried to override `corfu--make-buffer`, but it didn't work. Any suggestion on this? Thank you..

```
(defun zw/corfu--make-buffer (content)
  "Create corfu buffer with CONTENT."
  (let ((fr `((default (:family ,zw/font-fixed-pitch :height ,zw/font-size-fixed-pitch) default)))
        (ls line-spacing)
        (buffer (get-buffer-create " *corfu*")))
    (with-current-buffer buffer
      ;;; XXX HACK install mouse ignore map
      (use-local-map corfu--mouse-ignore-map)
      (dolist (var corfu--buffer-parameters)
        (set (make-local-variable (car var)) (cdr var)))
      (setq-local face-remapping-alist (copy-tree fr)
                  line-spacing ls)
      (cl-pushnew 'corfu-default (alist-get 'default face-remapping-alist))
      (let ((inhibit-modification-hooks t)
            (inhibit-read-only t))
        (erase-buffer)
        (insert content)
        (goto-char (point-min))))
    buffer))
(advice-add 'corfu--make-buffer :override
            'zw/corfu--make-buffer)
```

PR #165, opened 2022-05-10
Fix corfu get wrong size for frame in HiDPI screen, use fit-frame-to-buffer-1 adjust frame automatically.

manateelazycat:
Hi

I'm author of [lsp-bridge](https://github.com/manateelazycat/lsp-bridge), lsp-bridge's global is become fastest LSP client for Emacs.

I found corfu get wrong size for frame when corfu running in HiDPI screen.

Below is screenshot before patch:
2022-05-10 17-45-39 的屏幕截图

Below is screenshot after patch:
2022-05-10 17-45-52 的屏幕截图

We should use fit-frame-to-buffer-1 adjust frame automatically.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -449,7 +449,7 @@ The completion backend can override this with
       (set-window-buffer win buffer)
       ;; Mark window as dedicated to prevent frame reuse (#60)
       (set-window-dedicated-p win t))
-    (set-frame-size corfu--frame width height t)
+    (fit-frame-to-buffer-1 corfu--frame nil nil nil nil nil)
     (if (frame-visible-p corfu--frame)
         ;; XXX HACK Avoid flicker when frame is already visible.
         ;; Redisplay, wait for resize and then move the frame.


minad:
Thanks! This addresses #149. I am not sure if I trust fit-frame-to-buffer to do the right thing. Sometimes these resizing functions don't produce the expected result. I will try this for a while. The problem is also that the wrong height computation will break the child frame positioning.

minad:
How does lsp-bridge compare to eglot, lsp-mode, nox?

manateelazycat:
> How does lsp-bridge compare to eglot, lsp-mode, nox?

its the fastest one, hahaha.

manateelazycat:
> How does lsp-bridge compare to eglot, lsp-mode, nox?

lsp-bridge use python threading technologic to parse message from LSP server, Emacs don't need wait lsp-bridge.

lsp-bridge will compare LSP message with newest changed ticker, it will drop LSP message if message has expired.

lsp-bridge will push completion items to Emacs if LSP message not expired, then call corfu completion frame **manually**.

BTW, can you help me review code https://github.com/manateelazycat/lsp-bridge/blob/26ba3d108fbeef00c6b1246d1a4f318284a3855f/lsp-bridge.el#L380 ?

I turn off corfu-auto in lsp-bridge-mode, above code use for popup corfu frame manually, have better way to refresh corfu completion items?

Thanks.

minad:
What you could do alternatively is to write a completion-at-point-function which is more efficient and only returns if a valid result is available. The you could use `corfu-auto=0`. Did you consider this approach?

manateelazycat:
> What you could do alternatively is to write a completion-at-point-function which is more efficient and only returns if a valid result is available. When the Capf is called, the lsp server is contacted, nil is returned. When the capf is called and a result is available, return it. Furthermore you could show asynchronously incoming results via `corfu--auto-complete`, I mean forcing Corfu to show the display. Then you avoid replicating the code in your project.

I don't know how to change my capf function to async: https://github.com/manateelazycat/lsp-bridge/blob/dfdbd53c4cee4c1ce83822009c9cb497a675bcef/lsp-bridge.el#L394 

Because lsp-bridge will push result to Emacs, elisp side haven't async function wait lsp-bridge return.

Anyway, I push push new commit to use corfu--auto-complete instead copy code from corfu, https://github.com/manateelazycat/lsp-bridge/commit/dfdbd53c4cee4c1ce83822009c9cb497a675bcef 

Thanks for tips! ;)

manateelazycat:
@minad This patch is work for you?

minad:
I didn't have time to explore this yet, but it is unlikely. See my comment https://github.com/minad/corfu/pull/165#issuecomment-1122617611.

manateelazycat:
> I didn't have time to explore this yet, but this is unlikely to work. See my comment [#165 (comment)](https://github.com/minad/corfu/pull/165#issuecomment-1122617611).

Maybe you can add below code in corfu to fix this issue:
```elisp
 ;; Adjust default font height when running in HiDPI screen.
       (when (> (frame-pixel-width) 3000)
         (custom-set-faces '(corfu-default ((t (:height 1.3))))))
```

minad:
I tend to avoid such workarounds. It would be better to fix the font size computation, see #149 for the issue which tracks this.

Issue #147, opened 2022-03-25
Corfu-narrow: Kind-Narrowing


jdtsmith:
Got the idea from a post requesting this for company. It would be interesting to make it possible to narrow candidates by kind using a key binding or even with a small prompt. So like consult-narrow, with user-configured setup, wrapping (or setting) the predicate.  I would definitely use this to switch the display of functions vs. variables, for example.

minad:
Doesn't sound like a bad idea. It is quite easy to implement this by adjusting the completion predicate. However I see a number of practical problems:

* company-kind is defined by the backend, the set of kinds seems seems unrestricted.
* There are too many possible kinds.
* I doubt that the feature is very useful. It is nice to have for exploration, but I doubt it would pull its weight.

I don't want to blow up Corfu with more such features. The main feature of Corfu is that it is simple, easy to configure, easy to add to a configuration. I'd rather like to move out certain features, e.g., `corfu-show-location` and `corfu-show-documentation`. We could added narrowing as an extension. For now I added your proposal to the extension wishlish (#103). 

Anyway we don't have to take the lead here. The situation is like this: Company is defining its extensions the way it wants it and even adds them to core. I don't think Corfu will take off any time soon. If you look at the gh code search, company-mode is very widely used while nearly nobody uses corfu-mode.

Issue #173, opened 2022-05-24
LSP: corfu displays only one candidate for overloaded methods


oqwo:
Hello, 

First things first, thank you for this wonderful package. I really like how it manages to stay focused on completion and emacs integration.

I'm using eglot + JDTLS and corfu, and sometimes there are missing candidates. Today, I realized that when there are several overloaded candidate methods, only one of them is proposed.

Looking at `corfu--recompute-candidates`, I noticed a suspicious (?) call to `delete-consecutive-dups`
Without this call, overloaded methods do get displayed. 

There is probably a reason for deleting dups, but if this issue is really a bug, at least it may be a starting point for further investigations.

Thanks again!

oqwo:
Sorry, I didn't mentionned it but I did cross check against the default completion UI and company. Both propose all the overloaded candidates.

minad:
* ://github.com/emacs-mirror/emacs/blob/5f6e64a316adc6cd6ea140453718160643c6cd93/lisp/minibuffer.el#L1559
* https://github.com/emacs-mirror/emacs/blob/5f6e64a316adc6cd6ea140453718160643c6cd93/lisp/minibuffer.el#L2041

minad:
Why do you want to see duplicates anyway? To see the function signatures of the overloaded methods? Why does it make a difference for you when completing? Since the strings are equal the end result will be equal.

minad:
Also company deletes duplicates: https://github.com/company-mode/company-mode/blob/783287526a791590bea4f9a23992c2bfebdb4c8d/company.el#L1491

However it turns out that it only deletes them when the annotations are different! This is quite inefficient!

oqwo:
Thanks for your time and your explanations.

There is no doubt you know a lot more about the emacs internals than I do (I don't). I can just provide a screenshot of what happens on my side with emacs 28.1 without any customization + eglot master when I press C-M-i :

![image](https://user-images.githubusercontent.com/86486877/170124502-80888395-cd4f-4477-843f-0c7672adfd65.png)

Otherwise, you're right about me thinking that showing the signature of overloaded methods could be a plus. When working on some code base with poor/non-existent documentation, seeing the signature up-front can provide very useful hints. 
Moreover, while I don't care, eglot provide candidate signature expansion through yasnippet. You must select the appropriate candidate first, though

But maybe all of this is out the scope corfu targets?

[...]

PR #175, opened 2022-05-26
Add `corfu-toggle-kind-aggregation`, allows sorting by kind.

lorniu:
I often want to associate other completion items of the same kind (variables/functions) from the completion candicates. I searched and found somebody already asked for this (#147, #103).

I don't think `narrow` is a good way, instead re-sort by kind maybe a better choise.

So I implement the command `corfu-toggle-kind-aggregation` (bind to `M-r`). With this, I can switch between `normal order` and `kind-aggregation order` easily, this completely meets my needs.

Maybe somebody else need this, so I submit this PR. Free to change the code if you want to accept it, cause I don't very familiar with the completion/corfu API, it maybe not perfect.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -603,6 +603,11 @@ A scroll bar is displayed from LO to LO+BAR."
       (corfu--metadata-get 'display-sort-function)
       corfu-sort-function))
 
+(defun corfu--sort-candicates (candicates)
+  "Sort CANDICATES according the appropriate rule."
+  (delete-consecutive-dups
+   (funcall (or (corfu--sort-function) #'identity) candicates)))
+
 (defun corfu--recompute-candidates (str pt table pred)
   "Recompute candidates from STR, PT, TABLE and PRED."
   (pcase-let* ((before (substring str 0 pt))
@@ -623,7 +628,7 @@ A scroll bar is displayed from LO to LO+BAR."
     ;; this filtering, since this breaks the special casing in the
     ;; `completion-file-name-table' for `file-exists-p' and `file-directory-p'.
     (when completing-file (setq all (corfu--filter-files all)))
-    (setq all (delete-consecutive-dups (funcall (or (corfu--sort-function) #'identity) all)))
+    (setq all (corfu--sort-candicates all))
     (setq all (corfu--move-prefix-candidates-to-front field all))
     (when (and completing-file (not (string-suffix-p "/" field)))
       (setq all (corfu--move-to-front (concat field "/") all)))
--- a/extensions/corfu-kind-grouping.el
+++ b/extensions/corfu-kind-grouping.el
@@ -1,0 +1,71 @@
+;;; corfu-kind-grouping.el --- Sorting by king grouping for Corfu -*- lexical-binding: t -*-
+
+;; Copyright (C) 2022  Free Software Foundation, Inc.
+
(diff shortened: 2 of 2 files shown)


minad:
Instead of changing the candidate computation you could just implement a function corfu-sort-by-kind which can then be used as alternative corfu-sort-function.

lorniu:
Not the same,  `toggle-kind-aggregation` 1) sort based on the order provide by the backend 2) can easily switch back the origin order. A new sort function can not do these well, IMO.发自我的手机-------- 原始邮件 --------发件人： Daniel Mendler ***@***.***>日期： 2022年5月27日周五 03:18收件人： minad/corfu ***@***.***>抄送： lorniu ***@***.***>, Author ***@***.***>主    题： Re: [minad/corfu] Add `corfu-toggle-kind-aggregation`, allows sorting by kind. (PR #175)
Instead of changing the candidate computation you could just implement a function corfu-sort-by-kind which can then be used as alternative corfu-sort-function.

minad:
You can use the corfu-sort-override-function and set it to the to a new sort function based on the default sort function (or the backend sort function) and the kind grouping. Further modifications are not required.

lorniu:
I know `corfu-sort-override-function` is used to customize sorting rule.

But the case is that when I (or somebody else) config `corfu-sort-override-function` for some personal sorting rule (that is not kind-grouping),
and I also want to narrow/widden base on that.

> You can use the corfu-sort-override-function and set it to the to a new sort function 

If I implement like this, the custom sorting rule will be overrided. It's not convenient.

minad:
> But the case is that when I (or somebody else) config corfu-sort-override-function for some personal sorting rule (that is not kind-grouping), and I also want to narrow/widden base on that.

Then you simply compose that personal sorting rule with your narrowing sort function. Function composition is convenient. :)

I am not fond of adding an ad-hoc sorting/grouping mechanism on top since we already have `corfu-sort-override-function` which is meant for exactly this purpose. The `corfu-sort-override-function` gives you full flexibility.

lorniu:
What a pity, for me, I prefer this way.

But code as extension is acceptable?

minad:
No. I only find this acceptable as separate sorting function. I added the sort override mechanism for this purpose.

lorniu:
Maybe you don't fully understand what I mean.

I doesn't matter, but can you please extract `(delete-consecutive-dups (funcall (or (corfu--sort-function) #'identity) all))` from `corfu--recompute-candidates`, so that advices or other hacks can be more convenient?

minad:
> Maybe you don't fully understand what I mean.

I understand fully what you mean. I would appreciate a different approach from your side. The discussion went like this:

- You: Here I have this PR which adds ad-hoc kind grouping.
- I: Please try to use the sort function instead of adding an ad-hoc mechanism on top.
- You: I cannot do this because I want to compose functions.
- I: Then why not compose functions?
- You: Oh a pity. But I prefer this! Then take the extension!
- I: No, thanks.
- You: You probably don't understand me.

> I doesn't matter, but can you please extract (delete-consecutive-dups (funcall (or (corfu--sort-function) #'identity) all)) from corfu--recompute-candidates, so that advices or other hacks can be more convenient?

No. See https://github.com/minad/corfu/issues/173 about the duplicate discussion. The issue of duplicates is more complicated and so far unsolved. The completion API is under specified in this way.

PR #178, opened 2022-05-29
Add corfu-doc

galeo:
Hi @minad 

According to your request, I refactored the code in `corfu.el` that creates the child frame and sets the frame position and size. I also made some modifications to meet the needs of `corfu-doc`. As suggested by Stefan's code review, the version of `corfu-doc` is set to 0.8, also to transition from my repo version.

Please review the code when you have time.

Thanks.


minad:
I took a first look at your PR. Can you please turn on package-lint and fix the warnings, add missing docstrings, etc.?

minad:
@galeo I merged your modifications to corfu.el and made a few minor changes.

- https://github.com/minad/corfu/commit/450cf9d9e77906a986f82d6f1d409b5fc66e9175
- https://github.com/minad/corfu/commit/0bc9262131cc16d1938f3a93056273d919d22e2b

Can you please update corfu-doc.el accordingly? Then I will proceed with a more detailed review of corfu-doc.el.

minad:
There were two things I disagreed with, the REDISPLAY-HACK argument and the CONTENT-HANDLER. Can you explain why the REDISPLAY-HACK flag is needed for corfu-doc?

galeo:
> The content handler is unnecessary since you can just modify the buffer after `corfu--make-frame` given the refactoring I made.

I'll have a look and update `corfu-doc.el`.

> Can you explain why the REDISPLAY-HACK flag is needed for corfu-doc?

When this flag is on, the code `(sleep-for 0.01)` will cause a delay to the move doc frame.

galeo:
> Can you please update corfu-doc.el accordingly?

Done.

minad:
> Done.

Great thanks! I'll take another look at the code.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+  :safe #'integerp)
+
+(defcustom corfu-doc-resize-frame t
+  "Non-nil means resize the corfu doc frame automatically.
+
+If this is nil, do not resize corfu doc frame automatically."
+  :type 'boolean
+  :safe #'booleanp)
+
+(defcustom corfu-doc-display-within-parent-frame t
```
minad, line 82: This option should be removed until we have sorted out the support situation. I propose the following:

You extract a function `corfu-doc--display-area` from the position calculation function. This function returns the frame area by default. On Mac you can then advise override this function with a function which returns the display area. By doing that we still have an easy way to upgrade.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+  (setq corfu-doc--cf-popup-edges cf-popup-edges)
+  (setq corfu-doc--cf-window window))
+
+(defun corfu-doc-fetch-documentation ()
+  "Fetch documentation buffer of the current completion candidate."
+  (cond
+    ((= corfu--total 0)
+     (user-error "No candidates"))
+    ((< corfu--index 0)
+     (user-error "No candidate selected"))
```
minad, line 152: Return nil in those cases.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+    (list x y cf-doc-frame-width cf-doc-frame-height)))
+
+(defun corfu-doc--get-candidate ()
+  "Get the current completion candidate."
+  (and (> corfu--total 0)
+       (nth corfu--index corfu--candidates)))
+
+(defun corfu-doc--get-cf-popup-edges ()
+  "Get coordinates of the corfu popup."
+  (frame-edges corfu--frame 'inner))
```
minad, line 297: Do we really need these tiny one-line functions? I usually prefer if such small functions are inlined except if we want to advise them or if they contain some complicated logic. `corfu-doc--get-candidate` and `corfu-doc--get-cf-popup-edges` can both be inlined.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+            (frame-pixel-width corfu-doc--frame)
+            (frame-pixel-height corfu-doc--frame)))
+    (setq corfu-doc--cf-popup-edges (corfu-doc--get-cf-popup-edges))))
+
+(defun corfu-doc--update-popup (doc)
+  "Update the documentation popup with the DOC content."
+  (let ((doc-buffer
+          (corfu--make-buffer " *corfu-doc*" doc)))
+    (with-current-buffer doc-buffer
+      (visual-line-mode 1))
```
minad, line 325: Instead of enabling the visual-line-mode, I would prefer if the relevant variables are set directly. This makes it more transparent what is going on and we already set many variables via corfu--buffer-parameters.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+  :group 'corfu
+  (cond
+    (corfu-doc-mode
+     (corfu-doc--manual-popup-show)
+     (advice-add 'corfu--popup-show :after #'corfu-doc--popup-show)
+     (advice-add 'corfu--popup-hide :after #'corfu-doc--popup-hide))
+    (t
+     (corfu-doc--popup-hide)
+     (advice-remove 'corfu--popup-show #'corfu-doc--popup-show)
+     (advice-remove 'corfu--popup-hide #'corfu-doc--popup-hide))))
```
minad, line 401: I am not happy with the advice-based integration, since we have both manual-popup-show and popup-show. Please try the post-command-hook-based approach, I proposed.

1. Add a hook to the completion-in-region-mode.
2. In the completion-in-region-mode-hook check if corfu is enabled add a pch.
3. In the pch make all the necessary checks of the corfu popup state and selected candidate, and then show the doc popup, maybe timer based.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+  "Transition when showing or updating the doc popup."
+  (when (corfu-doc--popup-visible-p)
+    (if (and corfu-doc-mode corfu-doc-auto)
+        (when (and (not (null corfu-doc-delay)) (> corfu-doc-delay 0))
+          (if (> corfu-doc-delay corfu-doc-hide-threshold)
+              (corfu-doc--make-popup-invisible)
+            ;; clear buffer and update popup position immediately
+            (corfu-doc--clear-buffer)
+            (when (corfu-doc--cf-popup-edges-changed-p)
+              (corfu-doc--refresh-popup))))
```
minad, line 427: Instead of clearing the popup I would preserve the content of the last candidate, since this will avoid one flashing step. If users don't like this they can use a lower corfu-doc-hide-threshold.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+  "Determine whether the corfu popup is visible."
+  (and (frame-live-p corfu--frame)
+       (frame-visible-p corfu--frame)))
+
+(defun corfu-doc--should-show-popup (&optional candidate-index)
+  "Determine whether the doc popup should be showed.
+
+The optional CANDIDATE-INDEX is the the current completion candidate index,
+it should be compared with the value recorded by `corfu--index'."
+  (and (and (fboundp 'corfu-mode) corfu-mode)
```
minad, line 343: fboundp should not be necessary since you require corfu?

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+                                (- cf-parent-frame-height 1 1 y))
+                        (list (- display-width 1 lfw rfw 1 x cf-parent-frame-x)
+                              (- display-height 1 1 y cf-parent-frame-y)))))
+          (when (< width-remaining cf-doc-frame-width)
+            (setq cf-doc-frame-width
+                  (* (floor (/ width-remaining cw)) cw)))
+          (when (< height-remaining cf-doc-frame-height)
+            (setq cf-doc-frame-height
+                  (* (floor (/ height-remaining lh)) lh))))))
+    (list x y cf-doc-frame-width cf-doc-frame-height)))
```
minad, line 288: The computation is a bit complicated to follow. Can you decompose this slightly? As I suggested above (in my comment about corfu-doc-display-within-parent-frame), the display area computation should be extracted to a separate function. Maybe other sub-computations could be extracted too?

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+    (when (functionp function)
+      (with-selected-frame corfu-doc--frame
+        (with-current-buffer cf-doc-buf
+          (apply function args))))))
+
+(defun corfu-doc--popup-scroll (n)
+  "Scroll text of the documentaion buffer window upward N lines.
+
+See `scroll-up' for details."
+  (corfu-doc--funcall #'scroll-up n))
```
minad, line 461: Why is this a separate function? It could go into corfu-doc-scroll-up?

minad (changes requested):

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+                 (number :tag "seconds")))
+
+(defcustom corfu-doc-hide-threshold 0.2
+  "Threshold value to hide the documentation popup when browsing candidates.
+
+When the selected candidate is changed, if the value of `corfu-doc-delay'
+is greater than this threshold value, the documentation popup frame will
+be hided immediately. Else, just clear the doc frame content."
+  :type 'float
+  :safe #'floatp)
```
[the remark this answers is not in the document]
galeo: Done.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+          (if (corfu-doc--should-refresh-popup candidate)
+              (corfu-doc--refresh-popup)
+            (corfu-doc--popup-transition)
+            (when (and corfu-doc-mode corfu-doc-auto)
+              (setq corfu-doc--auto-timer
+                    (run-with-timer corfu-doc-delay nil
+                     #'corfu-doc--manual-popup-show corfu--index)))))
+      (corfu-doc--popup-hide))))
+
+(defun corfu-doc--funcall (function &rest args)
```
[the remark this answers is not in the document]
galeo: Done.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+  (advice-add 'corfu--popup-hide :after #'corfu-doc--cleanup)
+  (if (corfu-doc--popup-visible-p)
+      (progn
+        (corfu-doc--popup-hide)
+        (advice-remove 'corfu--popup-show #'corfu-doc--popup-show))
+    (corfu-doc--manual-popup-show)
+    (advice-add 'corfu--popup-show :after #'corfu-doc--popup-show)))
+
+;;;###autoload
+(define-obsolete-function-alias 'toggle-corfu-doc-mode #'corfu-doc-mode "0.7")
```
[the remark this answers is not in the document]
galeo: Done.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+               (res
+                ;; fix showing candidate location when fetch helpful documentation
+                (save-excursion
+                  (let ((inhibit-message t)
+                        (message-log-max nil))
+                    (funcall fun (nth corfu--index corfu--candidates))))))
+         (let ((buf (or (car-safe res) res)))
+           (with-current-buffer buf
+             (buffer-string)))
+       (user-error "No documentation available")))))
```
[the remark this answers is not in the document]
galeo: Done.

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
galeo: I think `corfu-doc--get-candidate` can be extracted to make the code clearer.  The `corfu-doc-terminal`

extensions/corfu-doc.el
galeo: I thought about setting the corresponding variables directly before, but I thought it was too detailed. Here is the equivalent code.

```elisp
(setq-local line-move-visual t)
(setq-local truncate-partial-width-windows nil)
(setq truncate-lines nil
      word-wrap t
      fringe-indicator-alist (list (cons 'continuation nil)))
```

Is this ok?

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
galeo: Done. `corfu-doc--funcall` is inlined.

extensions/corfu-doc.el
minad: Yes, setting them explicitly is better.

extensions/corfu-doc.el
minad: > [corfu-doc-terminal](https://codeberg.org/akib/emacs-corfu-doc-terminal) requires the corfu-doc--get-cf-popup-edges function and has an override advice on it.

Please don't take corfu-doc-terminal into account for now. The goal is to have corfu-doc in a form which is good enough for the Corfu repository. It is okay to break compatibility with this package since it is very new. Furthermore corfu-doc-terminal should be improved with better abstractions. The way it overrides a million functions is not a good idea. See https://codeberg.org/akib/emacs-corfu-doc-terminal/src/branch/master/corfu-doc-terminal.el#L537-L600.

Just compare, for corfu-terminal the situation is *MUCH BETTER*. https://codeberg.org/akib/emacs-corfu-terminal/src/commit/7c5a8a1c07b6c1a41b358b083d5bf7773701d26b/corfu-terminal.el#L204-L220

minad: Furthermore if you take a look at corfu-doc-terminal it has 500 lines. Compare this with the 200 lines of corfu-terminal. There is really something off here. It could even be that it is better to reimplement corfu-doc-terminal entirely without relying on corfu-doc. But hopefully we can find better abstractions to allowing hooking in corfu-doc-terminal in better ways. As it is now, it is just a recipe for breakage.

galeo: `corfu-doc--get-cf-popup-edges` has been inlined.

galeo: > I think `corfu-doc--get-candidate` can be extracted to make the code clearer. Maybe this function should be provided by corfu.

What do you think?

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
galeo: I've added a new option `corfu-doc-transition` to set how the transition should be.

extensions/corfu-doc.el
minad: > What do you think?

I don't think it makes sense to add this small function to Corfu, since Corfu wouldn't profit from it. For comparison, Vertico comes with a similar function in order to support Embark and Consult integration, but the Vertico function does not return the plain candidate but the *full candidate*. See https://github.com/minad/vertico/blob/b5119f6adfea40d2925d4488e9a7da4b4f333b17/vertico.el#L729-L743.

Let's not change anything in Corfu for now. I am a bit more conservative with regards to additions to the core package.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+  (setq corfu-doc--cf-window window))
+
+(defun corfu-doc--get-doc ()
+  "Get the documentation for the current completion candidate.
+
+The documentation is trimmed.
+Returns nil if an error occurs or the documentation content is empty."
+  (when-let
+      ((doc
+        (ignore-errors
```
minad, line 162: Why is there an ignore-errors?

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+                  (* (floor (/ width-remaining cw)) cw)))
+          (when (< height-remaining cf-doc-frame-height)
+            (setq cf-doc-frame-height
+                  (* (floor (/ height-remaining lh)) lh))))))
+    (list x y cf-doc-frame-width cf-doc-frame-height)))
+
+(defun corfu-doc--get-candidate ()
+  "Get the current completion candidate."
+  (and (> corfu--total 0)
+       (nth corfu--index corfu--candidates)))
```
minad, line 301: This function can be inlined.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+(defun corfu-doc--popup-scroll (n)
+  "Scroll text of the documentaion buffer window upward N lines.
+
+See `scroll-up' for details."
+  (when-let ((cf-doc-buf
+              (and (corfu-doc--popup-visible-p)
+                   (get-buffer " *corfu-doc*"))))
+    (with-selected-frame corfu-doc--frame
+      (with-current-buffer cf-doc-buf
+        (apply #'scroll-up n)))))
```
minad, line 465: This function can also be inlined since it is used only once.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+  (when (corfu-doc--popup-visible-p)
+    (if (and corfu-doc-mode corfu-doc-auto)
+        (when (and (not (null corfu-doc-delay)) (> corfu-doc-delay 0))
+          (pcase corfu-doc-transition
+            ('clear
+             (corfu-doc--clear-buffer)
+             (corfu-doc--refresh-popup))
+            ('hide (corfu-doc--make-popup-invisible))
+            (_ (corfu-doc--refresh-popup))))
+      (corfu-doc--popup-hide))))
```
minad, line 437: What's the difference between `corfu-doc--make-popup-invisible` and `corfu-doc--popup-hide`? Probably these functions should be merged?

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+     (advice-remove 'corfu--popup-hide #'corfu-doc--popup-hide))))
+
+(defun corfu-doc--popup-visible-p ()
+  "Determine whether the doc popup is visible."
+  (and (frame-live-p corfu-doc--frame)
+       (frame-visible-p corfu-doc--frame)))
+
+(defun corfu-doc--make-popup-invisible ()
+  "Make the doc popup invisible."
+  (make-frame-invisible corfu-doc--frame))
```
minad, line 419: This function should be removed or merged with `corfu-doc--popup-hide`.

extensions/corfu-doc.el
minad: Okay, that's good.

extensions/corfu-doc.el
minad: Not that it matters much in this case, but using overly small functions in Elisp is a recipe to make everything slower. My guideline is to extract functions if they reach a certain complexity (not the case here), or if I want to advise them (not the case here) or if extraction saves a lot of code duplication and reduces the overall amount of code (also not the case here, since the function is only used twice or three times I think).

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+         (cf-parent-frame (frame-parent corfu--frame))
+         (cf-frame--pos (frame-position corfu--frame))
+         (cf-frame-x (car cf-frame--pos))  ;; corfu--frame x pos
+         (cf-frame-y (cdr cf-frame--pos))
+         (cf-frame-width (frame-pixel-width corfu--frame))
+         (cf-frame-height (frame-pixel-height corfu--frame))
+         (cf-parent-frame-pos
+           ;; Get inner frame left top edge for corfu frame's parent frame
+           ;; See "(elisp) Frame Layout" in Emacs manual
+           (cl-subseq (frame-edges cf-parent-frame 'inner) 0 2))
```
minad, line 202: The cl-subseq seems unnecessary here and can be removed.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+documentation content, they can also be specified by optional parameters
+FWIDTH and FHEIGHT."
+  (let* (x y
+         ;; space between corfu popup and corfu doc popup
+         ;; set -1 to share the border
+         (space -1)
+         (cf-parent-frame (frame-parent corfu--frame))
+         (cf-frame--pos (frame-position corfu--frame))
+         (cf-frame-x (car cf-frame--pos))  ;; corfu--frame x pos
+         (cf-frame-y (cdr cf-frame--pos))
```
minad, line 196: It seems to me that you could also shorten this code by using `pcase-let`, e.g., directly destructure `cf-frame--pos` or `cf-parent-frame-pos`.

minad:
@galeo I just tested the current version again. It works very well now with the addition of corfu-doc-transition? Don't you prefer the behavior like this too? Maybe we could just get rid of corfu-doc-transition? I usually tend to avoid adding configuration options if it is not really necessary. Packages get more complex over time, so it is better to start smaller.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+(defun corfu-doc--calc-popup-position (&optional fwidth fheight)
+  "Calculate doc popup position (x, y), pixel width and height.
+
+The pixel width and height of the doc popup are calculated by the
+documentation content, they can also be specified by optional parameters
+FWIDTH and FHEIGHT."
+  (let* (x y
+         ;; space between corfu popup and corfu doc popup
+         ;; set -1 to share the border
+         (space -1)
```
minad, line 192: There is one more small issue regarding the border. I think Corfu should also work with border widths /= 0. I haven't tried this for a while. In corfu.el I retrieve the border width like this:

https://github.com/minad/corfu/blob/0bc9262131cc16d1938f3a93056273d919d22e2b/corfu.el#L478

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+  "Clear the doc popup buffer content and hide it."
+  (when (frame-live-p corfu-doc--frame)
+    (make-frame-invisible corfu-doc--frame)
+    (corfu-doc--clear-buffer)
+    (corfu-doc--set-vars nil nil nil)))
+
+(defun corfu-doc--cleanup ()
+  "Cleanup advices and hide the doc popup."
+  (advice-remove 'corfu--popup-hide #'corfu-doc--cleanup)
+  (unless corfu-doc-mode
```
minad, line 393: I get a variable warning due to this variable. You can either define the mode in the beginning or predeclare `(defvar corfu-doc-mode)`. But then I wonder why the check is even there. I still thin that the advice code must be cleaned up, since you use corfu-doc--cleanup and corfu-doc--popup-hide as advices, which I find hard to understand/follow. See my comment https://github.com/minad/corfu/pull/178/files#r889666846.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+                           (message-log-max nil))
+                       (funcall fun (nth corfu--index corfu--candidates))))))
+                 (let ((buf (or (car-safe res) res)))
+                   (with-current-buffer buf
+                     (buffer-string)))
+               nil))))))  ;; No documentation available
+    (unless (string-empty-p (string-trim doc))
+      doc)))
+
+(defun corfu-doc--calc-popup-position (&optional fwidth fheight)
```
minad, line 183: I tested corfu-doc a bit longer and I observed the slightly annoying behavior that the doc popup sometimes jumps from the left to the top to the right when scrolling, depending on if the documentation fits into the space. Wouldn't it be better to compute a *preferred direction* and then always place the popup there. The preferred direction would always be the direction in which we have the most free space. Then we would avoid the jumping. Do you think this would make the computation more complicated or maybe even simpler?

extensions/corfu-doc.el
```diff
@@ -0,0 +1,496 @@
...
+      (setq-local truncate-partial-width-windows nil)
+      (setq truncate-lines nil
+            word-wrap t
+            fringe-indicator-alist (list (cons 'continuation nil))))
+    (setq corfu-doc--frame
+          (corfu--make-frame
+           corfu-doc--frame corfu-doc--frame-parameters doc-buffer))
+    (apply #'corfu--move-frame
+           corfu-doc--frame
+           (corfu-doc--calc-popup-position))))
```
minad, line 340: Regarding #189 - Here we could get along with only a single call to corfu--make-frame.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,496 @@
...
+               (cf-popup-edges (frame-edges corfu--frame 'inner)))
+      (if (corfu-doc--should-refresh-popup candidate)
+          (corfu-doc--refresh-popup)
+        ;; fetch documentation and show
+        (if-let* ((doc (corfu-doc--get-doc)))
+            (progn
+              (corfu-doc--update-popup doc)
+              (corfu-doc--set-vars
+               candidate cf-popup-edges (selected-window)))
+          (corfu-doc--popup-hide))))))
```
minad, line 374: Regarding #189 - Here you have some logic which decides if the popup should be "refreshed" or "updated". This branching is unnecessary. Instead the popup should always be "updated". `corfu--make-frame` already takes care that the popup creation process is fast. This way we avoid two functions `corfu--make-frame` and `corfu--move-frame`.

extensions/corfu-doc.el
galeo: The popup frame should first be created with the document content. Only then can its size be calculated. When setting the popup frame resizable, `fit-frame-to-buffer` is used to make it fit to the content. Therefore, it needs to be processed in two steps. After the frame is created, it can be reused, and it only needs to adjust its size and position according to its updated content, which also makes it necessary to extract a separate function for processing.

minad: I see. Maybe it would be better to avoid `fit-frame-to-buffer` altogether since then we resize the frame twice (once by `fit-frame-to-buffer` and once by `corfu--move-frame`). Anyway - do you intent to address the other issues I mentioned? In particular the control flow should be overhauled (post-command-hook, state update, popup update). This will also help porting the package to the terminal later on.

galeo: > Instead we should first compute the size (`window-text-pixel-size`) and then create the frame. `window-text-pixel-size` can be called independently of the child frame.

I didn't know this function before. At first glance, I have a question how to get the WINDOW parameter passed to it?

I've just rebased the code for #189, will you merge it? Or do I keep refactoring the code against the current head of main branch?

minad: > I didn't know this function before. At first glance, I have a question how to get the WINDOW parameter passed to it?

One can probably use the current window and create a temporary buffer. There is a lot of work going on in the vtable.el package in Emacs master where pixel sizes of text are computed regularly. One can adopt the approach taken there.

> I've just rebased the code for https://github.com/minad/corfu/pull/189, will you merge it? Or do I keep refactoring the code against the current head of main branch?

No, I won't merge this for now. I care a lot about not breaking Corfu master and our last try didn't go so well. Also as I said, I prefer if we don't split the child frame function like this.

galeo: > One can probably use the current window and create a temporary buffer. There is a lot of work going on in the vtable.el package in Emacs master where pixel sizes of text are computed regularly. One can adopt the approach taken there.

I'll have a look. Thanks.

> No, I won't merge this for now. I care a lot about not breaking Corfu master and our last try didn't go so well. Also as I said, I prefer if we don't split the child frame function like this.

Can you continue to adjust the code so that the code for creating a buffer (`corfu--make-buffer`) and a frame (`corfu--make-frame`) can be reused by corfu-doc? Or do I need to adjust the code in the this pull request?

minad: I just mean a version which works for you. It doesn't matter if the branch diverged from the main branch. The interface to corfu.el will be small in the end anyway (`corfu--make-frame`). We discuss a bit too much here, I hope to see some progress. :)

extensions/corfu-doc.el
galeo: Hi @minad 

I still can't get better at showing the documentation popup by adding a post-command hook. In `corfu--atuo-complete`, `corfu--setup` is called before `corfu--update`. The candidates popup is not displayed when completion-in-region-mode is enabled. The hook of completion-in-region-mode still doesn't know when the corfu popup will be shown. A pch can be called with a timer, but the SECS parameter needs to be set to an appropriate value.

Here is a simple code that works, can you help with a review first? Thanks.

```elisp
(defun corfu-doc--post-command ()
  "Update documentaion popup after last command."
  (when (and (frame-live-p corfu--frame)
             (frame-visible-p corfu--frame))
    (if-let ((candidate (corfu-doc--get-candidate)))
        (progn
          (when corfu-doc--auto-timer
            (cancel-timer corfu-doc--auto-timer)
            (setq corfu-doc--auto-timer nil))
          (if (corfu-doc--should-refresh-popup candidate)
              (corfu-doc--refresh-popup)
            (corfu-doc--popup-transition)
            (setq corfu-doc--auto-timer
                  (run-with-timer
                   corfu-doc-delay nil
                   #'corfu-doc--popup-show corfu--index))))
      (corfu-doc--popup-hide))))

(defun corfu-doc--setup ()
  "Setup Corfu-doc."
  (if completion-in-region-mode
      (progn
        (run-with-timer 0.1 nil #'corfu-doc--post-command)
        (add-hook 'post-command-hook #'corfu-doc--post-command))
    (remove-hook 'completion-in-region-mode-hook #'corfu-doc--setup)
    (corfu-doc--teardown)))

(defun corfu-doc--teardown ()
  "Teardown corfu-doc."
  (remove-hook 'post-command-hook #'corfu-doc--post-command)
  (corfu-doc--popup-hide))

(defun corfu-doc--auto-post-command ()
  "Post command hook which initiates documentation popup."
  (when (and (not completion-in-region-mode)
             (not defining-kbd-macro)
             (not buffer-read-only)
             (corfu--popup-support-p))
    (add-hook 'completion-in-region-mode-hook #'corfu-doc--setup)))
    
(add-hook 'post-command-hook #'corfu-doc--auto-post-command 'append 'local)
```

minad: This looks as if it is going in the right direction. However I meant adding pch in the completion-in-region-mode-hook.

extensions/corfu-doc.el
galeo: This option has been removed.

extensions/corfu-doc.el
galeo: This calculation of the display area has been simplified.

extensions/corfu-doc.el
galeo: I used to get the documentation provided by the **helpful** package by overriding `elisp--company-doc-buffer`, it requires error handling. So, I hope this can be left as is.

minad: No, it cannot stay as is. If you want to override the function, the error handling should take place there. I am very hesitant to add superfluous `ignore-errors` in packages since these shadow bugs in other packages or in backends. Also they hurt debuggability. It is much better to fix the errors at the root. Then we end up with a system which is overall more robust.

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
galeo: Removed.

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,502 @@
...
+             (corfu--popup-support-p)
+             (frame-live-p corfu--frame)
+             (frame-visible-p corfu--frame)
+             (or (null candidate-index)
+                 (equal candidate-index corfu--index)))
+    (when-let ((candidate
+                (and (> corfu--total 0)
+                     (nth corfu--index corfu--candidates))))
+      (if (corfu-doc--should-move-popup candidate)
+          (corfu-doc--move-popup)
```
minad, line 355: Is it possible to remove `corfu-doc--move-popup` entirely? Instead just execute the `else` branch here, where the frame is recreated/moved. `corfu--make-frame` will make sure that this is efficient. I don't see a point in a separate move function.

extensions/corfu-doc.el
galeo: Fixed.

extensions/corfu-doc.el
minad: I think we have a different idea on how the display update should happen. I want the simplest thing possible - only a single `corfu--make-frame` function which updates the popup. Every time you want to move or update the popup in any way, call `corfu-doc-popup-show`, which then calls `corfu--make-frame`. No other manipulation of the popup should happen anywhere, except hiding it. This is exactly the same design as in Corfu itself. There are two functions `corfu--popup-show` (to show, update or move the popup) and `corfu--popup-hide`. Corfu-doc shouldn't be different.

extensions/corfu-doc.el
galeo: Taking into account the size of the parent frame, the way it's calculated now is to prioritize placing the doc popup below or above the candidates popup. When the bottom and top space is insufficient, consider placing it on the right or left side of the candidates popup. The specific calculation depends on the size of the area.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,496 @@
...
+  "Update the position of the doc popup when corfu popup edges changed."
+  (unless (corfu-doc--popup-visible-p)
+    (make-frame-visible corfu-doc--frame))
+  (when (corfu-doc--cf-popup-edges-changed-p)
+    (apply #'corfu--move-frame
+           corfu-doc--frame
+           (corfu-doc--calc-popup-position
+            (frame-pixel-width corfu-doc--frame)
+            (frame-pixel-height corfu-doc--frame)))
+    (setq corfu-doc--cf-popup-edges (frame-edges corfu--frame 'inner))))
```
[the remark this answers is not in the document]
galeo: Done.

extensions/corfu-doc.el
minad: That's great, but maybe the prioritized side should be configurable? Intuitively I would prioritize placing the popup on the right side, for left to right text, since displays these days are usually wide, 16:9 or wider. Why did you chose placing the popup below the Corfu popup?

extensions/corfu-doc.el
galeo: The code has been updated to directly use `corfu--make-frame`. However, `corfu--make-buffer` and `corfu--make-frame` have been modified so that they can be reused.

minad: > The code has been updated to directly use corfu--make-frame. However, corfu--make-buffer and corfu--make-frame have been modified so that they can be reused.

Good, thanks! I've only looked briefly but this seems to be the approach I would have chosen.

extensions/corfu-doc.el
galeo: Unlike corfu, corfu-doc often does not need to update the content of the frame, just adjust its position, especially when quickly browsing candidates. `corfu--make-frame` does more work than creating the frame, and adjusting the position and size may be more frequent for corfu-doc. I originally thought that separating the move-popup function might make the code a bit cleaner. 

I'll try to refactor this later.

minad: > Unlike corfu, corfu-doc often does not need to update the content of the frame, just adjust its position, especially when quickly browsing candidates.

I don't see how this is true. When you change the candidate (during scrolling or when filtering), the popup must be updated completely. Anyway, this is a premature optimization which makes everything more complicated without obvious benefits. `corfu--make-frame` is already optimized to avoid unnecessary duplicate expensive work.

My main goal here is to simplify corfu-doc as much as possible, since we will profit from that in the long term: Robustness, maintainability, portability for corfu-terminal etc.

extensions/corfu-doc.el
galeo: Got it. This has been updated.

extensions/corfu-doc.el
galeo: > I don't see how this is true. When you change the candidate (during scrolling or when filtering), the popup must be updated completely.

According to the default values of `corfu-doc-transition` and `corfu-doc-delay`, the content of the doc popup will not be updated when quickly browsing the candidates, only its position and size.

> My main goal here is to simplify corfu-doc as much as possible, since we will profit from that in the long term: Robustness, maintainability, portability for corfu-terminal etc.

I see. Thanks.

extensions/corfu-doc.el
galeo: This has been updated.

extensions/corfu-doc.el
minad: > According to the default values of corfu-doc-transition and corfu-doc-delay, the content of the doc popup will not be updated when quickly browsing the candidates, only its position and size.

Makes sense, but then you could still update the frame via `corfu--make-frame`. The only thing which won't be updated is the buffer, so you wouldn't necessarily call `corfu--make-buffer`.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+  "Corfu doc minor mode."
+  :global nil
+  :group 'corfu
+  (cond
+    (corfu-doc-mode
+     (corfu-doc--popup-show)
+     (add-hook 'completion-in-region-mode-hook #'corfu-doc--setup nil 'local))
+    (t
+     (corfu-doc--teardown)
+     (remove-hook 'completion-in-region-mode-hook #'corfu-doc--setup 'local))))
```
minad, line 480: local doesn't look right here. This is a global mode!°

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+                   #'corfu-doc--popup-show corfu--index))))
+      (corfu-doc--popup-hide))))
+
+(defun corfu-doc--setup ()
+  "Setup Corfu-doc."
+  (if completion-in-region-mode
+      (if corfu-doc-auto
+          (progn
+            ;; FIXME: can not show doc popup for the preselected candidate
+            (run-with-timer 0.1 nil #'corfu-doc--post-command)
```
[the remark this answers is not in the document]
minad: I will figure this out when I try your package the next time. Overall this looks almost good to go, except for the popup moving function which I would like to avoid.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,501 @@
...
+(defun corfu-doc--display-area-horizontal (width height)
+  "Calculate the horizontal display area for the doc popup.
+
+The WIDTH and HEIGHT of the doc popup are in pixels.
+
+The calculated area is in the form (X Y WIDTH HEIGHT)."
+  (pcase-let*
+      (;; space between corfu popup and corfu doc popup
+       ;; set -1 to share the border
+       (space -1)
```
minad, line 225: Did you try borders > 1?

extensions/corfu-doc.el
galeo: `space` is set to `(-border)` to share the border. I've tested when the border width is 0 or >1 and it works fine.

extensions/corfu-doc.el
galeo: See https://github.com/minad/corfu/pull/178#discussion_r897223980.

extensions/corfu-doc.el
galeo: corfu-doc-mode is now a local minor mode.

extensions/corfu-doc.el
galeo: Done.

extensions/corfu-doc.el
galeo: The code has been refactored. `corfu-doc--popup-show` is the only function to show, update or move the  doc popup.

extensions/corfu-doc.el
galeo: Outdated. See https://github.com/minad/corfu/pull/178#discussion_r896030017.

extensions/corfu-doc.el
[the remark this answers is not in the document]
galeo: > Overall this looks almost good to go, except for the popup moving function which I would like to avoid.

The popup moving function has been deleted.

> I will figure this out when I try your package the next time.

Thanks.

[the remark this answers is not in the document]
galeo: @minad Could you consider providing a hook with a name like `corfu-candidate-selected-hook`? This might make things easier.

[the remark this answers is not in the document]
minad: I didn't have much time to look into your PR recently. I let you know soon. What I've seen so far looks okay, so maybe a hook isn't necessary.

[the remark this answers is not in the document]
galeo: ping?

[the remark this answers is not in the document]
minad: I haven't forgotten your PR, but I still don't have much time for another while. I can only do a little bit of Emacs package maintainance these days. Next big thing will be to ensure that all my packages are compatible with Emacs 29.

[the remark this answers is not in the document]
galeo: OK. Thanks.

galeo:
@minad 

https://github.com/minad/corfu/blob/3490a42c30e4e32fe2d76dfc7736b649d37950db/corfu.el#L1193-L1194

Can these two lines of code be swapped? Render the candidates popup first and then enable `completion-in-region-mode`, so that the last FIXME entry in corfu-doc can be resolved.

minad:
> Can these two lines of code be swapped? Render the candidates popup first and then enable completion-in-region-mode, so that the last FIXME entry in corfu-doc can be resolved.

I don't think so.

galeo:
> > Can these two lines of code be swapped?
> 
> I don't think so.

According to the above two lines of code, `completion-in-region-mode` is first enabled, and then `corfu--update` renders the candidates popup. As you suggested earlier, by using the `completion-in-region-mode` hook,  it doesn't know when the corfu popup will be shown(with `corfu-preselect-first` set to `t`). This leads to the current problem that the corresponding documentation cannot be displayed for the preselected 1st candidate. 

So I think of the following 2 solutions:
- Render the candidates popup and then enable `completion-in-region-mode`, so that the documentation corresponding to the currently selected candidate can be displayed directly through the `completion-in-region-mode` hook and pch.

- Add a hook that runs every time the candidate is selected. corfu-doc can register its own function on
this hook. This might be a simpler/cleaner solution.

minad:
> This timer will be canceled in |corfu-doc--post-command|. I just tested
> it and it works fine. So I think the last issue blocking the merge has
> been resolved, can you please review it again?

Thanks! Using a timer doesn't sound like a great solution but maybe it
is good enough. I will take a closer look when I find time.

In the meantime, can you please fix all the typos in the docstrings?

galeo:
> In the meantime, can you please fix all the typos in the docstrings?

Sure. I've fixed the found typos as much as possible. I tried to get it right in the documentation, but since I'm not a native speaker there might be issues. If there is an inappropriate statement, please point it out or you can modify it directly.

Thanks.

extensions/corfu-doc.el
```diff
@@ -0,0 +1,500 @@
...
+(defcustom corfu-doc-delay 0.1
+  "The number of seconds to wait before displaying the documentation popup.
+
+The value of nil means no delay."
+  :type '(choice (const :tag "never (nil)" nil)
+                 (const :tag "immediate (0)" 0)
+                 (number :tag "seconds")))
+
+(defcustom corfu-doc-transition nil
+  "The method to transition the documentaion popup when browsing candidates.
```
minad, line 56: typo

galeo: Fixed.

galeo:
> Maybe this could also help in avoiding the timer solution you mentioned in [#178 (comment)](https://github.com/minad/corfu/pull/178#issuecomment-1277350605)? You can call corfu--update to ensure that the Corfu state is initialized.

While it is possible to call `corfu--update` to ensure the Corfu state is initialized, it still doesn't know when the corfu candidates popup will show up. The candidates popup must be displayed before the documentation popup is displayed. So I don't think the previous problem can be solved by calling `corfu--update`. Or maybe I just haven't figured it out yet.

> I am sorry that you have to develop against such a moving target, but getting the core logic right is more important than adding extensions. I hope it is not too hard to adapt corfu-doc.

That's okay, I also think it's more important that the core logic is correct.

> Another thing which came to my mind recently: We already have the corfu-info extension which shows the documentation and definition in a separate buffer. If we add corfu-doc here it may make sense to use consistent names?
> 
>     * corfu-doc -> corfu-doc-popup
> 
>     * corfu-info -> corfu-doc-buffer
> 
> 
> What do you think?

I think yes, it's better to use consistent names. Do you prefer renaming after merging corfu-doc or before?
After the merge, I consider deprecating the old corfu-doc in my repo and removing it from melpa. Do you think it is okay? Or you can make some suggestions.

minad:
> While it is possible to call corfu--update to ensure the Corfu state is initialized, it still doesn't know when the corfu candidates popup will show up. The candidates popup must be displayed before the documentation popup is displayed. So I don't think the previous problem can be solved by calling corfu--update. Or maybe I just haven't figured it out yet.

Okay sure. I just thought I mention it, in case it could help with the solution. When I review this closer I will also think again about the problem.

> I think yes, it's better to use consistent names. Do you prefer renaming after merging corfu-doc or before? 

I would prefer to rename the extensions as part of this PR, but the renaming of corfu-info should be done in a separate commit. What about the names corfu-docup and corfu-docwin?

> After the merge, I consider deprecating the old corfu-doc in my repo and removing it from melpa. Do you think it is okay? Or you can make some suggestions.

Yes that sounds okay. My recommendation would be to replace corfu-doc temporarily on ELPA/MELPA with an almost empty package which only prints a warning (maybe as an autoload). A few months later it can be removed entirely.

galeo:
> I would prefer to rename the extensions as part of this PR. The renaming of corfu-info should be done in a separate commit as part of this PR. What about the names corfu-docup and corfu-docwin?

I think corfu-docup and corfu-docwin may not be clear enough and a bit redundant. Do you consider merging corfu-info into corfu-doc as one extension? Use a name corfu-doc to provide two documentation display methods, popup and buffer (window). 

> My recommendation would be to replace corfu-doc temporarily on ELPA/MELPA with an almost empty package which only prints a warning (maybe as an autoload). A few months later it can be removed entirely.

I will follow your suggestion. Thanks.

minad:
> I think corfu-docup and corfu-docwin may not be clear enough and a bit redundant. Do you consider merging corfu-info into corfu-doc as one extension? Use a name corfu-doc to provide two documentation display methods, popup and buffer (window).

No, I prefer to have them separate since the two extensions seem quite different, in both complexity and user interaction. Also they are mutually exclusive, you'd want to use either the one or the other.

galeo:
> No, I prefer to have them separate since the two extensions seem quite different, in both complexity and user interaction. Also they are mutually exclusive, you'd want to use either the one or the other.

If so, then I think the names corfu-docup and corfu-docwin are fine. 

> I would prefer to rename the extensions as part of this PR. The renaming of corfu-info should be done in a separate commit as part of this PR.

Is it possible to wait until corfu-doc is merged before renaming corfu-info? I think the renaming of corfu-info should be separate from this pull request. I fixed some issues I found, and I hope you can review the code again. Or do you want to rename corfu-doc to corfu-docup first and then review the code?

david-vicente:
Hi, I've been following this PR. I just wanted to say that

> * corfu-doc -> corfu-doc-popup
> 
> * corfu-info -> corfu-doc-buffer

sounds better to me than 

> corfu-docup and corfu-docwin

Thank you both for your work!

minad:
@david-vicente Okay, thanks for your input. We may go with the longer, more explicit name then. @galeo Can you please make the renaming as part of this PR? The corfu-info renaming should be a separate commit, while the corfu-doc renaming can be squashed with your other commits.

galeo:
@minad 

> * corfu-doc -> corfu-doc-popup
> 
> * corfu-info -> corfu-doc-buffer

The renaming of corfu-doc and corfu-info is done.

minad:
@galeo I just tried corfu-doc-popup again. It seems the documentation popup alignment has somehow gotten worse than I remember. When the documentation popup is placed to the right of the candidates popup, it isn't aligned with the candidates popup and there is a small gap. Furthermore if it is placed below there is a gap. Is this alignment intentional?

galeo:
> When the documentation popup is placed to the right of the candidates popup, it isn't aligned with the candidates popup and there is a small gap. Furthermore if it is placed below there is a gap. Is this alignment intentional?

No, this was not intentional. I think it should be a problem with Emacs, which handles the frame position inconsistently on different window systems. People has reported a [issue](https://github.com/galeo/corfu-doc/issues/23) in the NS port Emacs. I've also filed a [bug](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=58627) report upstream. I've observed the inconsistency across versions of X11/GTK, Mac port, and NS port. I don't know if there is something wrong with the way I understand and handle the child frame position. Currently, there should be no problem on the Mac port and the NS port. I'll find a device asap to fix the problem on X11/GTK.

extensions/corfu-doc-popup.el
```diff
@@ -0,0 +1,517 @@
...
+corfu doc mode is turned on and `corfu-doc-popup-auto' is set to Non-nil."
+  (interactive)
+  (when corfu-doc-popup-mode
+    (if (corfu-doc-popup--visible-p)
+        (progn
+          (corfu-doc-popup--teardown)
+          (remove-hook 'completion-in-region-mode-hook #'corfu-doc-popup--setup))
+      (corfu-doc-popup--show)
+      (corfu-doc-popup--setup)
+      (add-hook 'completion-in-region-mode-hook #'corfu-doc-popup--setup))))
```
AkibAzmain, line 455: Is this worth autoloading?  This is can't a entry point of the file, because it needs `corfu-doc-popup-mode` to be enabled to something.

extensions/corfu-doc-popup.el
```diff
@@ -0,0 +1,517 @@
...
+  "Scroll text of doc popup window upward N lines.
+
+If ARG is omitted or nil, scroll upward by a near full screen.
+See `scroll-up' for details."
+  (interactive "p")
+  (when-let ((cf-doc-buf (and (corfu-doc-popup--visible-p)
+                              (get-buffer " *corfu-doc-popup*"))))
+    (with-selected-frame corfu-doc-popup--frame
+      (with-current-buffer cf-doc-buf
+        (funcall #'scroll-up n)))))
```
AkibAzmain, line 417: Can you please put this part in another, non-interactive function?  Advising a function is saner than a command, and I need to `:around` advice it for terminal support (which is likely to never call the original function).

extensions/corfu-doc-popup.el
```diff
@@ -0,0 +1,517 @@
...
+                                           border border)
+                                        (- (frame-pixel-height corfu-doc-popup--frame)
+                                           border border))))))))
+        (if doc-updated-p
+            (corfu-doc-popup--set-vars candidate
+                                       (frame-edges corfu--frame 'inner-edges)
+                                       (selected-window))
+          (when cfp-edges-changed-p
+            (setq corfu-doc-popup--cf-popup-edges
+                  (frame-edges corfu--frame 'inner-edges))))))))
```
AkibAzmain, line 379: Can you please move the child-frame related code from here?  This function uses frames so much get I can't advice it.

AkibAzmain (changes requested):
Please address the issues I pointed out to allow using `corfu-doc-popup` on terminal.

extensions/corfu-doc-popup.el
galeo: Thanks. This has been resolved.

extensions/corfu-doc-popup.el
galeo: I remember that this seems to be a change requested by mind. He thinks that too many small functions should be avoided and that can be directly inlined if possible.

extensions/corfu-doc-popup.el
galeo: @minad has reviewed it and I have been adjusting this part of the code. This request has been waiting for a long time to be merged. I would like to hear minad's thoughts on whether it should continue to be adjusted and then wait for the time to merge. Or would you consider overriding this function directly?

minad: @galeo Please don't make changes now which facilitate integration with corfu-terminal. We can improve that in a second step.

minad:
@galeo I tried corfu-doc again and unfortunately the auto popup doesn't appear reliably for me. Sometimes it appears and after a while it stops appearing. Then I can manually summon the popup using the corfu-doc-popup-toggle, but it won't update anymore automatically. Did you observe this too?

The second issue is that the popup placement is a bit unpredictable. I'd like to always have the popup placed below, to the right or the left during a single completion session. I find it disturbing if the popup jumps around. My proposal is to compute the largest area next to the popup and always use that for the popup. You can choose some weights, e.g., such that the space to the left or right is rated higher.

These two issues should be addressed before I can merge this. I know you've waited a long time for this to get merged. I hope we can finally push it over the finish line!

minad:
@AkibAzmain 

> Please address the issues I pointed out to allow using corfu-doc-popup on terminal.

Please be patient regarding supporting corfu-doc-popup in corfu-terminal. I've tried corfu-doc a few times recently but unfortunately it still has some issues, which prevent the feature from being merged. We should focus first on getting corfu-doc ready, before we discuss how terminal support should be implemented.

galeo:
@minad 

> I tried corfu-doc again and unfortunately the auto popup doesn't appear reliably for me. Sometimes it appears and after a while it stops appearing. Then I can manually summon the popup using the corfu-doc-popup-toggle, but it won't update anymore automatically. Did you observe this too?

I haven't spent much time on Emacs recently due to work. I feel that it may be related to the previous problem.
Because it is not sure when the corfu popup is first displayed for the preselected item, a repeating timer is used. Here is the code in `corfu-doc-popup--setup`:
```elisp
      (if corfu-doc-popup-auto
          (progn
            (setq corfu-doc-popup--preselect-first-timer
                  (run-with-timer 0.1 0.1 #'corfu-doc-popup--post-command))
            (add-hook 'post-command-hook #'corfu-doc-popup--post-command
                      'append 'local))
```
The timer will be canceled in `corfu-doc-popup--post-command`:
```elisp
(when corfu-doc-popup--preselect-first-timer
      (cancel-timer corfu-doc-popup--preselect-first-timer)
      (setq corfu-doc-popup--preselect-first-timer nil))
```
I feel that this problem should be caused by the execution of the repeating timer. Is there be a problem that the timer is not actually canceled and causes `corfu-doc-popup--post-command` to be executed repeatedly? I need to find some time to reproduce this problem. Please help check here if you have time.

> The second issue is that the popup placement is a bit unpredictable. I'd like to always have the popup placed below, to the right or the left during a single completion session. I find it disturbing if the popup jumps around. My proposal is to compute the largest area next to the popup and always use that for the popup. You can choose some weights, e.g., such that the space to the left or right is rated higher.

The position of the doc popup is indeed set according to the calculated largest area. Here is the code:
```elisp
  (pcase-let*
      ((`(,width . ,height)  ;; popup inner width and height
         (corfu-doc-popup--size width height))
       (`(,v-x ,v-y ,v-w ,v-h)
         (corfu-doc-popup--display-area-vertical width height)))
    (if (and (>= v-h height) (>= v-w width))
        (list v-x v-y v-w v-h)
      (pcase-let ((`(,h-x ,h-y ,h-w ,h-h)
                    (corfu-doc-popup--display-area-horizontal width height)))
        (if (>= (* v-w v-h) (* h-w h-h))
            (list v-x v-y v-w v-h)
          (list h-x h-y h-w h-h)))))
```

Because corfu popup sometimes has a large width(especially some programming languages ​​with long completion items), and there is not enough space in the horizontal direction. So the vertical direction is first selected to display. The horizontal direction is selected only if there is not enough space in the vertical direction. If there is not enough space in the horizontal direction, the display position will be selected according to the calculated size. During a single completion session, the size of the corfu popup changes, resulting in a change in the size of the area available to display the doc popup. If the position of the popup is limited to a certain direction, it may cause insufficient display space. It is not difficult to adjust here, the above code is the `corfu-doc-popup--display-area` function. In addition, it may be related to personal habits, or this issue can be put on hold.

minad:
@galeo Thanks for looking into this. 

> I feel that this problem should be caused by the execution of the repeating timer.

I agree. I guess we have to find a better solution than the timer since it is too unreliable like this. Maybe it makes more sense to use an :after advice on corfu--exhibit and update the doc popup there according to the Corfu state. Did you have that before changing to post-command-hooks?

> During a single completion session, the size of the corfu popup changes, resulting in a change in the size of the area available to display the doc popup. If the position of the popup is limited to a certain direction, it may cause insufficient display space.

This doesn't sound right. I've even observed the jumping around when scrolling around. During scrolling the Corfu candidates popup size doesn't change. So this also has something to do with the size of the doc popup? 

Anyway, I find it disturbing enough that the Corfu candidate popup itself changes size and jumps around. The doc popup should stick to a direction. It should be sufficient to store the last direction and keep on using this direction even if the Corfu popup changes its size.

minad:
I just checked - in your original corfu-doc package you advice the corfu--popup functions, which I don't really like. But in contrast advising corfu--exhibit seems okay, since this is the main function which updates the Corfu state.

https://github.com/galeo/corfu-doc/blob/da931367802d01e87e1e496ba5b51aa9126a705d/corfu-doc.el#L484-L485

galeo:
@minad 

> I guess we have to find a better solution than the timer since it is too unreliable like this. Maybe it makes more sense to use an :after advice on corfu--exhibit and update the doc popup there according to the Corfu state.

> But in contrast advising corfu--exhibit seems okay, since this is the main function which updates the Corfu state.

I've updated the code to use an after advice for `corfu--exhibit`. Please help verify that the first issue is resolved.

> This doesn't sound right. I've even observed the jumping around when scrolling around. During scrolling the Corfu candidates popup size doesn't change. So this also has something to do with the size of the doc popup?

Yes. The size of the doc popup is calculated dynamically and its size may change when the current candidate changes. In addition, when scrolling, the size of the corfu popup will indeed change(e.g. when the length of the selected candidate exceeds the width of the popup).

> Anyway, I find it disturbing enough that the Corfu candidate popup itself changes size and jumps around. The doc popup should stick to a direction. It should be sufficient to store the last direction and keep on using this direction even if the Corfu popup changes its size.

I've also updated the code so that the doc popup will stick to a direction during a completion session. Please try again and see how it works.

Thanks.

minad:
@galeo Thanks! Sounds good.

extensions/corfu-doc-popup.el
AkibAzmain: Yeah, too many small functions (including lambda) result in inefficient, but provide more points add to advice.  However, don't that now, it's just a reminder, first make the code stable.

AkibAzmain:
@minad Yes, advices are often bad, since they often depend on the internals of a package.  `corfu-doc` (the version on MELPA) currently has a child-frame dependent part and a child-frame independent part, both clearly separated (like Corfu).  There are ten advices in `corfu-doc-terminal`, but each of them just override the corresponding function (I use `:around` to use the original function in GUI), replacing the whole child-frame dependent part with Popon equivalent (similar to a Company frontend).  I have tried my best to keep things simple, and these advices (and those in `corfu-terminal`) are actually some of the most sane advices I have ever seen (I don't like to praise my own code, but these advices are really better that others).   IIUC the advices will break if and only if the advised function's behavior change.  So maintaining the terminal UI shouldn't be hard if we cooperate.  (If you don't think so, you're welcome review the code, but please don't judge by number of advices.)  Thanks to @galeo for cooperating to implement `corfu-doc-terminal` (galeo/corfu-doc#12 and galeo/corfu-doc#13)!

minad:
@AkibAzmain 10 advices are way too much and I won't give you guarantees about this. If you look for example at corfu.el or vertico.el, I use two or three advices. The same applies to your corfu-terminal, which is fine.

AkibAzmain:
@minad I don't need guarantee.  I have no problem if you break it from time to time, as long as you keep child-frame dependent part clearly separated, so that I can override that.

minad:
@AkibAzmain Of course. But if we have to split up the code such that you can install ten advices, we won't do it. This is not a good design. I hope we can refactor the code somewhat that you don't need as many advices and such that the separation between popup and child frame code is more clear. I understand your sentiment about "good advices", which cleanly replace an implementation, but the pure number should still somewhat stay in check. 

Note that I did the same in corfu.el itself, where we have a separation between frame and popup. This separation is what enabled corfu-terminal in the first place. If we can do the same for corfu-doc-popup, great, if not, there is no other way than reinventing the wheel for the terminal :shrug:

extensions/corfu-doc-popup.el
```diff
@@ -0,0 +1,547 @@
...
+  (let* ((cw (default-font-width))
+         (lmw (* cw corfu-left-margin-width))
+         (rmw (* cw corfu-right-margin-width))
+         (fp (copy-alist corfu--frame-parameters)))
+    (setf (alist-get 'left-fringe fp) (ceiling lmw)
+          (alist-get 'right-fringe fp) (ceiling rmw))
+    fp)
+  "Default doc child frame parameters.")
+
+(defvar corfu-doc-popup--auto-timer nil
```
minad, line 104: Why do we still need the auto timer?

extensions/corfu-doc-popup.el
```diff
@@ -0,0 +1,547 @@
...
+                  (run-with-timer
+                   corfu-doc-popup-delay nil
+                   #'corfu-doc-popup--show nil corfu--index))))
+      (corfu-doc-popup--hide))))
+
+(defun corfu-doc-popup--cf-exhibit-after-advice (&rest _args)
+  "After advice for `corfu--exhibit'.
+To display the doc popup for the preselected completion candidate."
+  (when (and corfu--candidates (>= corfu--index 0))
+    (corfu-doc-popup--post-command)))
```
minad, line 506: It seems we have now both a post-command-hook and a vertico--exhibit advice? I meant that replacing the post-command-hook with *only an advice* should be sufficient.

minad:
Okay, I've decided to merge this now as is. We can still improve it as part of Corfu.

galeo:
@minad That's great, thank you!

AkibAzmain:
> your corfu-doc-terminal package is longer or as long as corfu-doc itself

Yes, because it has to show a buffer in a popon, which needs some code.  If I remove three biggest functions, line count decreases from 668 to 390, less than corfu-doc.  My code is only for showing the Popon, if I add the doc fetch code from `corfu-doc`, it would almost certain reach thousand lines.

minad:
@AkibAzmain I've reworked and simplified the extension heavily. Please take a look and let me know where we would have to make changes. The problem is that the extension is relying a lot on frame functions.

AkibAzmain:
@minad I've noticed that after installing the latest Corfu from elpa-devel.  I've just opened the issue #248 for this discussion, with two proposals to help implement terminal support.  I've also opened [akib/corfu-terminal#16 on Codeberg](https://codeberg.org/akib/emacs-corfu-terminal/issues/16).

And quoting from [one of my emails](https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg00897.html) to emacs-devel list:

> Before someone asks me to consider assigning my copyright, I would like
> to say that a printed copy of assignment agreement is in my hands right
> now, but I can't sign it now, and I don't know when I'll able to sign
> (don't get me wrong, I really want to sign it).

AkibAzmain:
Looks like you have renamed the extension, just noticed now.

AkibAzmain:
@minad,

> It would be best if you can create a PR. Is there a chance that you can get the copyright assignment?

I can't create a merge-able PR now, but I can indeed create a PR for you (or anyone else who completed the paperwork, perhaps @galeo) to take inspirations from make the real changes.  Should I try to do this?

AkibAzmain:
I think you got me wrong, that PR wouldn't be for merging, that would be for someone (with the papers, of course) to take some inspiration from and _do the real change themselves_.

minad:
@AkibAzmain Look, this won't be an approach which is legally sound. End of discussion.

AkibAzmain:
@minad Thanks for the clarification.  EOD.

PR #183, opened 2022-06-01
Add option to specify the maximum suffix width.

JimDBh:
This allows us to truncate the candidates when they are very long,
so the suffix can always be shown.


minad:
I don't want to truncate the candidate in favor of the annotation suffix. The candidate is always more important than the annotation.

minad:
But feel free to add an advice of corfu--format-candidates to the wiki. It is easy to do such customizations on the level of the user configuration. You could then even activate this only for certain backends.

Issue #186, opened 2022-06-06
Leftover height when narrowing candidates


elken:
![image](https://user-images.githubusercontent.com/2872862/172104412-2ccce536-1504-4cd8-9185-5c32ac35eaec.png)

Screenshot above describes the issue; when initiating completion, the frame is populated fine. However, once I type a few more characters to narrow the selection, the "old" height still persists. Maybe #149 as I make `'corfu-default` inherit from `'tooltip`?. There are only 2 valid candidates, so I would expect the height to reflect that.

I'm not sure if this is a bug or a feature, if it's the former how simple would it be to add a toggle for that behaviour?

[config](https://github.com/elken/doom/tree/master/modules/completion/corfu)

minad:
Ouch! These child frame resizing issues should be sorted out by now. Can you reproduce this in emacs -Q? Which emacs version is this? Pgtk? Which Corfu commit?

minad:
I think it is this ugly bug: https://github.com/minad/corfu/issues/17

minad:
Please check if the issue occurs before: https://github.com/minad/corfu/commit/a262a214325c7e33704bb1fade0fa40042ff93a5

This could be a regression due to @galeo's recent corfu-doc related refactorings

elken:
> Can you reproduce this in emacs -Q? 

I went to try it on macos and it works as expected here, which "helps".

> Which emacs version is this? 

28.1

> Which GUI toolkit? 

X11

> Which Corfu commit?

corfu is unpinned so whatever the latest commit is

> Please check if the issue occurs before: https://github.com/minad/corfu/commit/a262a214325c7e33704bb1fade0fa40042ff93a5

I will do this after work tonight along with `emacs -Q` and see what comes back :)

minad:
Please check the newest commit. This is probably just a regression.

elken:
> Please check the newest commit. This is probably just a regression. (should work on both emacs -Q and in Doom)

I will report back one way or another after testing, thanks as always minad :)

minad:
One more question - do you use gtk3/gnome/cinnamon? The aforementioned bug is specific to those gui toolkits.

elken:
> One more question - do you use gtk3 and (gnome or cinnamon)? The aforementioned bug is specific to those gui toolkits.

I do :) I edited my original comment a few times since posting it

elken:
This is indeed still an issue I'm afraid :( 
[...]

PR #189, opened 2022-06-08
Try to fix regression that the child frame not resize dynamically

galeo:
See #186.

The commit is based on 248137e1e03e885b708fb3bf4b340458f60a0b91. If it works, I'll rebase it to the head of the master branch.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -368,11 +368,11 @@ The completion backend can override this with
     map)
   "Ignore all mouse clicks.")
 
-(defun corfu--make-buffer (content)
-  "Create corfu buffer with CONTENT."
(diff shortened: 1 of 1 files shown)


elken:
Confirms this also fixes the issue.

Previous commit
image

d85edf2e794d70e94791ad8a8a8bc7d4df27f33f
image

minad:
Thanks @galeo! But now that I looked more into corfu-doc, I think we have to change the way corfu-doc is integrated. Why do you need two functions (make and move)? The idea of `corfu--make-frame` is that it takes over the whole child frame creation process. There shouldn't be a need for separate functions. I add another comment at the corfu-doc PR.

PR #206, opened 2022-07-26
README: Mention corfu-doc-terminal

jgarte:


minad:
This can be mentioned later. The situation around corfu-doc is in flux.

jgarte:
> This can be mentioned later. The situation around corfu-doc is in flux.

Where can I read about the flux?

@minad Do you use `corfu-doc`?

 I have stopped using it because I realized that I prefer to just jump to the definition or get any info in the minibuffer.

minad:
See #178. We plan to add corfu-doc here soon. Then corfu-terminal will include support directly and corfu-doc-terminal won't be needed.

jgarte:
Sounds great

PR #225, opened 2022-08-27
corfu--pre-command: insert candidate by unreading current command and using corfu-insert

aaronjensen:
Continuation/new version of #169 

This allows evil-repeat to work.

Rather than continuing discussion on an old, out-of-date PR, I figured I'd open this so you can see the latest (and what I've been using).

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -892,6 +892,13 @@ there hasn't been any input, then quit."
       (redisplay 'force)) ;; XXX HACK Ensure that popup is redisplayed
      (t (corfu-quit)))))
 
+(defun corfu--unread-this-command-keys ()
+  (when (> (length (this-command-keys)) 0)
+    (setq unread-command-events (nconc
+                                 (listify-key-sequence (this-command-keys))
+                                 unread-command-events))
+    (clear-this-command-keys t)))
+
 (defun corfu--pre-command ()
   "Insert selected candidate unless command is marked to continue completion."
   (when corfu--preview-ov
@@ -902,7 +909,8 @@ there hasn't been any input, then quit."
              ;; See the comment about `overriding-local-map' in `corfu--post-command'.
              (not (or overriding-terminal-local-map
                       (corfu--match-symbol-p corfu-continue-commands this-command))))
-    (corfu--insert 'exact)))
+    (corfu--unread-this-command-keys)
+    (setq this-command 'corfu-insert-exact)))
 
 (defun corfu-insert-separator ()
   "Insert a separator character, inhibiting quit on completion boundary.
@@ -1062,6 +1070,14 @@ Quit if no candidate is selected."
   (interactive)
   (if (>= corfu--index 0)
       (corfu--insert 'finished)
+    (corfu-quit)))
+
+(defun corfu-insert-exact ()
+  "Insert current candidate with the `exact' status.
+Quit if no candidate is selected."
+  (interactive)
+  (if (>= corfu--index 0)
+      (corfu--insert 'exact)
     (corfu-quit)))
 
 (defun corfu--setup ()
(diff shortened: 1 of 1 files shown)


minad:
Thanks. I appreciate the update. But to be clear - I remain unconvinced about the method. My skepticism won't change if you ask repeatedly about merging this.

aaronjensen:
Sorry, that wasn't my intent. I'll stop asking about it. I'm sorry for bothering you.

minad:
No worries! I appreciate your contributions and I don't feel bothered by it. I just don't want to give a wrong impression here. Since you asked in #169 how I want to proceed - I consider the issue only a minor issue, since it occurs only in combination with Evil, and also only in a subfeature of Evil. I think it is okay to wait and see about this issue. Corfu doesn't see wide adoption (yet?). There are (or were) much more pressing issues, e.g., compatibility issues with lsp-mode due to the lsp-mode Capf implementation.

It would be nice to have better support for Evil, but I am against adding complications here which do not have other positive effects or make the code simpler (even if it is only a minor thing like this). Evil is not a great reference package due to its quirks. I've gotten many reports regarding Evil across various packages, which required some special workarounds. Furthermore when I tried Evil I discovered that it had other quality issues (poorly made global advices, compilation issues, ...), such that I removed it again from my configuration. Some of the problems are understandable. It is certainly not easy to retrofit the Vim model onto Emacs, such that most of these details are ugly necessities.

Anyway, I think it is fair to ask if it would be possible to improve compatibility on the side of Evil such that Corfu would just work as is.

aaronjensen:
Sounds good. For anyone who comes to this and wants it to work, you can apply this after requiring corfu:

```el
  (defun corfu--unread-this-command-keys ()
    (when (> (length (this-command-keys)) 0)
      (setq unread-command-events (nconc
                                   (listify-key-sequence (this-command-keys))
                                   unread-command-events))
      (clear-this-command-keys t)))

  (defun corfu--pre-command ()
    "Insert selected candidate unless command is marked to continue completion."
    (when corfu--preview-ov
      (delete-overlay corfu--preview-ov)
      (setq corfu--preview-ov nil))
    (when (and (eq corfu-preview-current 'insert)
               (/= corfu--index corfu--preselect)
               ;; See the comment about `overriding-local-map' in `corfu--post-command'.
               (not (or overriding-terminal-local-map
                        (corfu--match-symbol-p corfu-continue-commands this-command))))
      (corfu--unread-this-command-keys)
      (setq this-command 'corfu-insert-exact)))

  (defun corfu-insert-exact ()
    "Insert current candidate with the `exact' status.
Quit if no candidate is selected."
    (interactive)
    (if (>= corfu--index 0)
        (corfu--insert 'exact)
      (corfu-quit)))
```

> Anyway, I think it is fair to ask if it would be possible to improve compatibility on the side of Evil such that Corfu would just work as is.

From what I understand about how evil-repeat works, I'm not sure what approach would be possible. I think ultimately, the fact that the modifications happen in a pre-command hook for every command is what prevents evil-repeat from knowing how to detect the changes and repeat them. The only thing they could do is ensure that all commands are marked needing to record buffer changes (which would be wrong for many commands) and then ensure that the buffer change recording included pre-commands (meaning that the evil-repeat hook would need to ensure that it was the very first pre-command hook).

I'm going to hold off on opening an issue on evil at the moment and just use the above patch. Maybe I or someone else will feel motivated to do so later.

kenranunderscore:
@aaronjensen Sorry to bother you, but I couldn't quite follow at some point. I've tried using the above by evaluating the expressions in a scratch buffer, but it didn't have the desired effect for me. Maybe my expectation was wrong though: I was in `*scratch*`, typing `macroe` and completing with some of `corfu`'s suggestions. Then I entered normal mode and pressed `.` to repeat the insertion, but all that was inserted was what I initially typed, `macroe`. Am I going wrong here somewhere?

aaronjensen:
@kenranunderscore hard to say. You probably need at least this:

```el
(mapc #'evil-declare-ignore-repeat
          '(corfu-next
            corfu-previous
            corfu-first
            corfu-last))

    (mapc #'evil-declare-change-repeat
          '(corfu-insert
            corfu-insert-exact
            corfu-complete))
```
 
Also, I've only tested w/ the "tab-n-go" type behavior in corfu.

kenranunderscore:
@aaronjensen Thank you very much! This actually makes it work :) This was the only and a major pain point in `corfu` for me until now.

aaronjensen:
I updated the patch above to account for the recent function rename.

ahan98:
@aaronjensen your work has not gone unnoticed! thanks a ton for your patches :)

aaronjensen:
For anyone following this, the workaround is now part of evil-collection: https://github.com/emacs-evil/evil-collection/pull/798

Issue #222, opened 2022-08-22
Preview single candidates similar to company


elken:
Kinda related to https://github.com/minad/corfu/issues/215 but I thought it should be a new issue.

Similar to [company](https://company-mode.github.io/manual/Frontends.html#Preview-Frontends), what would be required in order to add this behaviour to Corfu?

Specifically the ability to preview a single candidate inline with a customizable face(s)
![image](https://user-images.githubusercontent.com/2872862/185904274-8b1b7708-61da-4ba9-b0ce-86704fc02a70.png)
![image](https://user-images.githubusercontent.com/2872862/185904392-934fc1e0-2dd4-46c9-a68e-a879812d2727.png)

If this already exists and I've missed something feel free to tell me I'm an idiot and close this! :D

minad:
Duplicate #215 and #211. Feel free to implement this if you are interested and propose a PR. I don't think it is a particularly useful feature, especially if you use more flexible completion styles.

elken:
I agree on the not useful front, but someone asked for it earlier.

I may look into it, but maybe not. Have other things to look into for when I start looking at adding this as a doom module soon ™️ .

minad:
You are not the first to ask. The difficulty is to find a design which fits nicely with what we have right now. I hope there will be some good feedback from Doom since you have a full blown Company configuration.

We could preview the first candidate in the style of Company if it matches the prefix. Otherwise we could display something like `input<candidate>` for non-prefix matches. Given such preview one could even decide to not show the popup for unique candidates or only if you are scrolling down (`corfu--index>0`).

elken:
Lots of people are pretty excited about Corfu, so I would like to think I could provide a good enough experience! :D 

I think it's only truly useful when you have a single candidate and even then behind `corfu-preview-single-candidate` or something. If true and we have 1 candidate, display the inline preview rather than invoking the frame.

As you say, with orderless in the mix trying to handle that for every kind of candidate would be too tricky.

minad:
> As you say, with orderless in the mix trying to handle that for every kind of candidate would be too tricky.

Well, as I mentioned we could simply display `complicated-input|<the full candidate>` via the overlay, where | is the cursor. Not sure if this looks good for in-buffer completion, bit it would be similar to Ido or vertico-unobtrusive/flat in the minibuffer.

Issue #215, opened 2022-08-09
'corfu-preview-current' with 'corfu-preselect-first' doesn't preview first


david-vicente:
It feels a bit odd that the first entry is treated differently from the rest with regards to previewing.

Here `random` is not previewed next to the prompt
![image](https://user-images.githubusercontent.com/57544666/183540959-b5bd25d1-b7ea-4f9c-93fd-8a5fd416dc83.png)

but `arange` is previewed
![image](https://user-images.githubusercontent.com/57544666/183541088-6bd6f59e-4957-49fe-a57a-62e3f4659eee.png)

For example in Vscode every completion entry is previewed
![image](https://user-images.githubusercontent.com/57544666/183540907-5b320c31-c333-4fce-a964-92afaed3e93d.png)

Is this intended?

Thanks

minad:
This is intended. We cannot preview the a candidate without overwriting the prompt. The vscode preview is less general and only works since the candidate matches the prefix. Corfu also supports non-prefix completion styles like substring, flex and orderless.

Issue #211, opened 2022-08-02
Option to preview top candidate without popup


meain:
Company had an option to preview the top candidate in the buffer without opening a popup. In most cases, this is all I need and don't actually need the popup to show up; but if I need the all the options, I can just wait some more time and get it. This was much less intrusive in most cases where I just want to complete some long name that I know it will be the top candidate or I have completed this thing before and is in the top of my history or I know what exactly to prefix to get it. Is there some way I can recreate this with corfu?

https://user-images.githubusercontent.com/14259816/182438320-81c16679-6cc9-4d1a-b406-bf484373bef4.mp4

minad:
No, this feature doesn't exist yet in Corfu. I've considered before to add it but so far I've got along just fine without it. The reason is that previewing the top candidate does not generalize well. It only works well with prefix completion (the basic completion style), where the input matches the prefix of the candidate. I am a fan of non-prefix completion (orderless, substring, regexp, flex). For non-prefix completion the input doesn't match the candidate prefix and the preview of the topmost candidate is not possible, since the preview overlay would have to overwrite the entire input.

Anyway I am not opposed to adding this feature if someone is interested in implementing it.

meain:
Just for anyone running into this in the future, https://code.bsdgeek.org/adam/corfu-candidate-overlay/ seems to be doing something similar.

PR #232, opened 2022-09-20
Preview first candidate

minad:
This implements preview of the first candidate, similar to the proposals in #211, #215 and #222. We could even hide the popup in more cases, e.g., when there is only a single candidate. Maybe it also makes sense to hide the popup always as long as it hasn't been requested explicitly. See also the reddit discussion in https://old.reddit.com/r/emacs/comments/xi9jdx/autocomplete_previews_while_typing_w_plugable/ started by @ideasman42. But I am not sure if we can replicate the behavior described there satisfactorily given the existing Corfu interaction model.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -894,7 +894,8 @@ there hasn't been any input, then quit."
         (corfu--done str 'finished)))
      ;; 3) There exist candidates => Show candidates popup.
      (corfu--candidates
-      (corfu--preview-current beg end)
+      (when (or (not (cdr corfu--candidates)) (/= corfu--index corfu--preselect))
+        (corfu--preview-current beg end))
       (corfu--echo-documentation)
       ;; TODO unobtrusive mode
       ;; (when (or (not corfu-preview-current)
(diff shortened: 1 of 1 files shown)


ashton314:
I've just played with this for a minute—looks pretty! I'll have to tweak the config so the popup doesn't show up with only one candidate. I don't think this corfu style is my favorite—I'm a `(setq corfu-auto nil)` kind of guy—but I could see this being very nice for a lot of people.

JosephFerano:
Thank you for implementing this feature. I tried it out and visually it looks nice. However the input lag seems significantly slower compared to the normal popup. My use case is similar to #211, I am looking to have "fish-style" autocompletion where I only see the first candidate and instead use vertico for full autocompletion. If corfu could have the option to disable the popup and always show the preview, that would be amazing.

For the performance issue, I can help troubleshoot if needed.

I am running emacs 29 with pgtk, compiled with the following flags;

```
./configure --with-pgtk \
    --with-native-compilation \
    --without-compress-install \
    --with-imagemagick \
    --with-tree-sitter \
    --with-json \
    CFLAGS="-O2 -pipe -march=znver3 -fomit-frame-pointer"
make -j 15
```

And here is my config so far

```
(straight-use-package
 '(corfu :type git :host github :repo "minad/corfu" :branch "preview-first"))
(global-corfu-mode t)
(setq corfu-auto t)
(setq corfu-auto-delay 0.1)
(setq corfu-count 1)
(setq corfu-preview-current t)
```

minad:
@JosephFerano The lag is indeed a problem. Since we are showing the preview directly inline, the computation has much stronger timing requirements. Example:

1. Type `com` -> Display `com<pletion>`
2. Add another `p` -> Display `comp<pletion>` wouldn't be valid anymore

Of course one could use some heuristic which then drops the additional input from the candidate etc. But it will be a major complication and deviation from the current Corfu behavior. I don't plan to make this part of Corfu.

PR #243, opened 2022-11-08
Simplify the definition of the minor-mode

dr-scsi:
* extensions/corfu-indexed.el (corfu-indexed-mode): Use (if ...) instead of (cond ...) with two clauses for minor-mode definition.


minad:
Hmm, I just saw that I use cond rather consistently for all my mode definitions. Therefore I reverted this again. Tbh I think it is a bit of a pointless change.

dr-scsi:
@minad - It is a cosmetic change and probably more a matter of taste.  So let leave things the way they are and acc. to your taste 👍

PR #247, opened 2022-11-18
Don't use *Help* buffer for displaying docstring

twlz0ne:
Using `*Help*` buffer will interfere with people who are writing code while reading documentation in the Help window.

Steps to reproduce:

1. Start `emacs -Q -l test-corfu.el`.
1. Execute `M-x describe-function RET pcase RET` to open `*Help*` in other window.
1. Type `(corfu-` In `*scratch*` buffer.
1. Press `C-M-i` to active completion.
1. Wailting for the docframe to show.
1. The content of `*Help*` has changed.

### DIFF
--- a/extensions/corfu-docframe.el
+++ b/extensions/corfu-docframe.el
@@ -110,6 +110,10 @@ See `frame-edges' for details.")
   (and (frame-live-p corfu-docframe--frame)
        (frame-visible-p corfu-docframe--frame)))
 
+(defun corfu-docframe--advice-help-buffer ()
+  "Advice `help-buffer' to return the name of a buffer for displaying docstring."
+  (buffer-name (get-buffer-create " *Corfu-doc*")))
+
 (defun corfu-docframe--get-doc (candidate)
   "Get the documentation for CANDIDATE.
 Returns nil if an error occurs or the documentation content is empty."
@@ -117,7 +121,9 @@ Returns nil if an error occurs or the documentation content is empty."
               (res (save-excursion
                      (let ((inhibit-message t)
                            (message-log-max nil))
-                       (funcall fun candidate)))))
+                       (cl-letf (((symbol-function 'help-buffer)
+                                  #'corfu-docframe--advice-help-buffer))
+                         (funcall fun candidate))))))
     (with-current-buffer (or (car-safe res) res)
       (setq res (buffer-string)))
     (and (not (string-empty-p (string-trim res))) res)))


minad:
Thanks. I can reproduce this issue, but I think I have to treat it as "wontfix". This seems like a deficiency of the `:company-doc-buffer` API. If we open the doc buffer manually in a separate buffer as in `corfu-info-documentation` it seems desired that we override the help buffer. Of course for `corfu-docframe`/`company-quickhelp`, we wouldn't want the buffer to be overridden. The problem is that the fix you just proposed is backend-dependent and we cannot add workarounds for each and every backend here. Maybe @dgutov has some further insight on this issue and the API design of `:company-doc-buffer`? It would probably be best if `:company-doc-buffer` supports an optional buffer name argument? We could roll that out slowly. Frontends could call `:company-doc-buffer` with argument guarded by a `condition-case` and fallback to the zero argument call if the first call failed.

dgutov:
> If we open the doc buffer manually in a separate buffer as in corfu-info-documentation it seems desired that we override the help buffer.

I don't know if that's a given. For simplicity of the implementation, I would probably drop that requirement.

> It would probably be best if :company-doc-buffer supports an optional buffer name argument? We could roll that out slowly.

Please elaborate: which conditions it will be used in, and who will provide the buffer name?

minad:
@dgutov The API :company-doc-buffer is used for two things:

1. Display the help buffer in a separate window. This case works well, but will override the current help buffer.
2. Display the help buffer contents in a small popup (company-quickhelp, corfu-popupinfo). In this case the help buffer gets overridden unexpectedly, in particular if the help buffer is currently visible in a side window.

I've seen that there is also the quickhelp-string API call in Company, but a :company-quickhelp-string Capf extension function doesn't seem to exist. Also it is probably not implemented widely?

dgutov:
Suppose in case 1 the `*Help*` buffer already exists, and it shows a doc for a certain variable.

Then we enter completion, and ask Emacs to show a doc for a different variable, from the completions list. If `doc-buffer` reuses `*Help*` in this scenario, then it's going to show the info for the new variable, overriding the previous text. And it's going to keep showing it, instead of restoring to the previous state, when we press `C-g`. At least that is the case for `elisp-completion-at-point`, I'm not sure how it is for the implementation this issue is talking about.

So... even in case 1 we are behaving somewhat dirtily. Even though reusing the same window usually is handy, reusing the buffer itself is probably not ideal.

If we ensure that the returned buffer is always dedicated to completion docs (and not a common one like `*Help*`), we get a "cleaner" behavior in both cases. Reusing the window could be arranged through other means.

> I've seen that there is also the quickhelp-string API call in Company, but a :company-quickhelp-string Capf extension function doesn't seem to exist. Also it is probably not implemented widely?

`quickhelp-string` was originally devised to show a different version of the buffer in quickhelp, see https://github.com/company-mode/company-quickhelp/issues/28#issuecomment-180043936.

Indeed, it's not used very widely, and I sometimes forget it exists. But it could also be used to solve the main problem described here.

minad:
> If we ensure that the returned buffer is always dedicated to completion docs (and not a common one like *Help*), we get a "cleaner" behavior in both cases. Reusing the window could be arranged through other means.

That's probably an acceptable solution too. :company-doc-buffer could always return a special buffer, only for completion help.

On the other hand if the buffer is shown in a separate window, reusing the help buffer is not bad. It will feel more integrated with the usual Emacs help behavior.

dgutov:
> On the other hand if the buffer is shown in a separate window, reusing the help buffer is not bad. It will feel more integrated with the usual Emacs help behavior.

There is something to be said for the existing behavior, yes. It just has its wrinkles too, as described above.

PR #249, opened 2022-11-18
Limit printing depth/length to reduce doc lag

twlz0ne:
There is a lag when displaying doc of a variable which has a value of large list.

Steps to reproduce:

1. Start Emacs:

```sh
$ emacs -Q --eval "\
  (progn
    (toggle-debug-on-error)
    (require 'seq)
    (setq user-emacs-directory
          (car (seq-filter
                #'file-exists-p
                (list (format \"~/.emacs.d/%s/\" emacs-version)
                      (format \"~/.emacs.d/%s.%s/\" emacs-major-version emacs-minor-version)
                      \"~/.emacs.d/\"))))

    (setq package-user-dir (concat user-emacs-directory \"elpa/\"))
    (package-initialize)
    ;; Add addtional local package here:
    (add-to-list 'load-path \"~/repos/emacs-corfu\")
    (add-to-list 'load-path \"~/repos/emacs-corfu/extensions\")
    ;; ---------------------------------------------------------------------------
    (require 'corfu)
    (require 'corfu-docframe)
    (setq corfu-auto t)
    (setq corfu-quit-at-boundary t)
    (setq corfu-preview-current nil)
    (setq corfu-docframe-auto t)
    (setq corfu-docframe-delay 0)
    (global-corfu-mode 1)
    (corfu-docframe-mode 1))"
```

1. Type `load-`.
1. Active completion, and the first two candidates should be `load-path` and `load-history`.
1. Press `C-n`.
1. Press `C-n`.

### DIFF
--- a/extensions/corfu-docframe.el
+++ b/extensions/corfu-docframe.el
@@ -69,6 +69,16 @@
   :group 'corfu
   :type 'integer)
 
+(defcustom corfu-docframe-print-level 3
+  "Value for `print-level' while printing value in doc."
+  :group 'corfu
+  :type 'integer)
+
+(defcustom corfu-docframe-print-length 12
+  "Value for `print-length' while printing value in doc."
+  :group 'corfu
+  :type 'integer)
+
 (defcustom corfu-docframe-resize t
   "Resize the corfu doc popup automatically if non-nil."
   :group 'corfu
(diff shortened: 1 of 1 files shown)


minad:
Thanks! I've pushed your fix but without adding customization variables. Tbh I also consider this fix problematic, since it is also backend dependent. At least it is a little bit more generic since we just let bind the print variables.

minad:
Just fyi, the extension is still undergoing heavy churn. I even renamed it again to corfu-infoframe, to reflect that it can display both documentation and location.

PR #260, opened 2022-11-24
Fix compiler warnings

dr-scsi:
* extensions/corfu-popupinfo.el (corfu-popupinfo--display-area-horizontal)
(corfu-popupinfo--display-area-vertical)
(corfu-popupinfo--display-area): Fix warning about docstring has wrong usage of unescaped single quotes.

### DIFF
--- a/extensions/corfu-popupinfo.el
+++ b/extensions/corfu-popupinfo.el
@@ -250,7 +250,7 @@ in the form of (X Y WIDTH HEIGHT)."
 The WIDTH and HEIGHT of the info popup are in pixels.
 The calculated area is in the form (X Y WIDTH HEIGHT DIR).
 DIR indicates the horizontal position direction of the info popup
-relative to the candidate popup, its value can be 'right or 'left."
+relative to the candidate popup, its value can be \\='right or \\='left."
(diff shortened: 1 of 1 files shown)


minad:
Maybe we should better remove the single quotes. I find the escaping unreadable and ugly.

dr-scsi:
By [convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html), Lisp symbols should be written in docstrings like `` `this'``.   In order to prevent a hyperlink, one can put the word `symbol` before the symbol name.  So in this case, `` `vertical'``, `` `left'`` and `` `right'`` would be correct.  I didn't suppose them because I had the impression that you like the `'vertical` version which is also used in many packages.  But at the end, it's your call `:-)`

minad:
"Documentation tips"

minad:
Anyway, if you feel strongly we could use symbol `vertical' etc. Feel free to create a PR. I only want to avoid the quotation since this hurts the readability of the doc string itself.

dr-scsi:
I'm easy, really.  I'm just trying to help avoid compiler warnings for the packages I use.  So whatever fits for you, I'll take it 👍

PR #261, opened 2022-11-24
corfu-popupinfo: Make sure the popup has a minimum height of 1 line

galeo:
The attribute `:height` of `corfu-popupinfo` face is set to 0.8 by default, the calculated minimum height (1 line height) of the popup will be smaller than the actual minimum height of the popup (default 1 line height), resulting in a deviation in the position setting when the popup is in the top direction.

### DIFF
--- a/extensions/corfu-popupinfo.el
+++ b/extensions/corfu-popupinfo.el
@@ -219,9 +219,10 @@ all values are in pixels relative to the origin. See
 (defun corfu-popupinfo--size ()
   "Return popup size as pair."
   (let* ((cw (default-font-width))
+         (lh (default-line-height))
          (margin (* cw (+ (alist-get 'left-margin-width corfu-popupinfo--buffer-parameters)
                           (alist-get 'right-margin-width corfu-popupinfo--buffer-parameters))))
-         (max-height (* (default-line-height) corfu-popupinfo-max-height))
+         (max-height (* lh corfu-popupinfo-max-height))
          (max-width (* cw corfu-popupinfo-max-width)))
     (or (when corfu-popupinfo-resize
           (with-current-buffer " *corfu-popupinfo*"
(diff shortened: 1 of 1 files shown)


minad:
Wouldn't it be possible to instead lift the minimum height restriction of 1 for the popup?

minad:
I tried to lift the restriction, but for some reason this doesn't work.

galeo:
> Wouldn't it be possible to instead lift the minimum height restriction of 1 for the popup?

> I tried to lift the restriction, but for some reason this doesn't work.

@minad I tried that too, but it didn't work. I checked that the value of the `height` parameter passed to `set-frame-size` is correct, but the frame height is limited to 2 * border + default line height. Setting `frame-resize-pixelwise` to non-nil has no effect either.

I don't know if this restriction can be lifted. Here is a link I found about the description of the `set-frame-size` function: https://lists.gnu.org/archive/html/emacs-devel/2019-10/msg00487.html

> (1) Change the semantics of the PIXELWISE argument for a number of
> user functions.  For 'set-frame-size' we could say that if its
> value is 'linewise', then use the 'window-default-line-height'
> value of the frame's selected window.

Perhaps it has something to do with this?

minad:
You can create smaller frames, but...

https://github.com/minad/corfu/blob/1b74e981027e2413d518d6277e0ab642bc5497ab/extensions/corfu-popupinfo.el#L238-L242

galeo:
> You can create smaller frames, but...

Have a quick try and it is able to create a smaller frame, but the frame got cut off first, and it could return to normal after waiting a few seconds.

minad:
I think the behavior may also be platform dependent. It seems we have no other possibility than accepting a minimal size of 1 default line for now. Nevertheless we should open an issue on the Emacs bug tracker.

galeo:
I've submitted a bug report upstream [#59547](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=59547).

galeo:
You can add it. Or wait for the maintainer's reply before adding it.

minad:
Let's wait a bit.

PR #264, opened 2022-11-26
corfu-popupinfo: Support always-* display directions and minimum thresholds for the popup

galeo:
This pull request contains two commits. The first one attempts to support displaying the popup in the always-* directions. In the second commit, add the popup minimum width and height settings. Try to check the width and height of the popup when `corfu-popupinfo-direction` is set to 'horizontal or 'vertical', when it is less than the threshold, recalculate the popup display area. In this case, it may cause the direction of the popup to change, resulting in visual jumps, but it has been avoided as much as possible.

@minad It is only in draft status at present. You can take a look first and try to see if there are any mistakes. Thanks.

### DIFF
--- a/extensions/corfu-popupinfo.el
+++ b/extensions/corfu-popupinfo.el
@@ -75,12 +75,22 @@ popup can be requested manually via `corfu-popupinfo-toggle',
   :group 'corfu)
 
 (defcustom corfu-popupinfo-max-width 80
-  "The max width of the info popup in characters."
+  "The maximum width of the info popup in characters."
+  :type 'integer
+  :group 'corfu)
+
+(defcustom corfu-popupinfo-min-width 30
+  "The minimum width of the info popup in characters."
   :type 'integer
   :group 'corfu)
 
 (defcustom corfu-popupinfo-max-height 10
-  "The max height of the info popup in characters."
+  "The maximum height of the info popup in characters."
+  :type 'integer
+  :group 'corfu)
+
+(defcustom corfu-popupinfo-min-height 1
+  "The minimum height of the info popup in characters."
   :type 'integer
   :group 'corfu)
 
@@ -89,15 +99,17 @@ popup can be requested manually via `corfu-popupinfo-toggle',
   :type 'boolean
   :group 'corfu)
 
-(defcustom corfu-popupinfo-direction 'horizontal
-  "Preferred direction for the popup."
-  :type '(choice (const horizontal)
-                 (const vertical)
-                 (const always-horizontal)
-                 (const always-vertical)
-                 ;; TODO always-left and always-right are unsupported
-                 (const always-left)
-                 (const always-right))
+(defcustom corfu-popupinfo-direction '(right left vertical)
+  "Preferred directionse for the popup in order."
+  :type '(repeat
+          (choice
+           (const left)
+           (const right)
+           (const vertical)
+           (const always-horizontal)
+           (const always-vertical)
+           (const always-left)
+           (const always-right)))
   :group 'corfu)
 
 (defvar corfu-popupinfo-map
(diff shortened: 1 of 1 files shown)


minad:
Thanks for working on this. I think the logic is getting a bit out of hand. We should refactor first, simplifying the area computation. I think we can simplify greatly and cut away a lot of code.

First we compute the desired popup size `ps`. We should introduce four variables holding the areas at the top, bottom, left, right: `at`, `ab`, `al`, `ar`. These variables are always computed. Furthermore introduce the variables `av` and `ah`, where `ah` is `at` or `ab` depending on position of candidates popup and `ah` is the larger one of `al` and `ar`. Then check all the cases one by one.

1. Update `corfu-popupinfo--dir` if it is non-nil. Check if direction area is sufficient (larger than minimum size), otherwise reset `corfu-popupinfo--dir` to nil.
2. `always-left` or `corfu-popupinfo--dir=left` -> `al`
3. `always-right` or `corfu-popupinfo--dir=right` -> `ar`
4. `always-vertical` or `corfu-popupinfo--dir=vertical` -> `av`
6. `always-horizontal` -> if `ps` fits in `ar` use `ar`, otherwise `ah`
8. `horizontal` and `ps` fits in `ar` -> `ar`
8. `horizontal` and `ps` fits in `al` -> `al`
9. `horizontal` and `ah` larger than `av` -> `ah`
10. otherwise -> `av`

Does this make sense?

minad:
The way I wrote the pseudo code above already suggests that we should introduce helper functions:

- `corfu-popupinfo--area-larger-p`
- `corfu-popupinfo--fits-p`

galeo:
> Does this make sense?

Yes, really great. Thanks for the detailed pseudo code. My initial idea was to minimize existing code changes and reduce unnecessary computation. This leads to code that is not particularly clear and complex in logic.

> We should introduce four variables holding the areas at the top, bottom, left, right: `at`, `ab`, `al`, `ar`. These variables are always computed.

Is it necessary to calculate all of them all the time? Although they may be used in subsequent conditional branches, I don't think it is necessary to calculate each time. Of course this leads to more clarity and less code.

In addition, one thing I want to say is that under certain conditions, there will be no `ps` larger than `ms`, how to deal with this situation?

Would you like to implement the code yourself? This will be better.

minad:
On 11/27/22 12:15, galeo wrote:
> Is it necessary to calculate all of them all the time? Although they may
> be used in subsequent conditional branches, I don't think it is
> necessary to calculate each time. Of course this leads to more clarity
> and less code.

Yes, it is not particularly efficient to compute the variables all the
time. The pseudo code above is how one would write something in Haskell.
But the computation won't be the bottleneck here. We already compute the
vertical and horizontal display area in the current code. I would really
go with the simplest solution here for clarity.

> In addition, one thing I want to say is that under certain conditions,
> there will be no |ps| larger than |ms|, how to deal with this situation?

In this case the natural thing should happen. We fall through to the
last "otherwise" branch and then use the larger area (av or ah). This is
the best we can do.

> Would you like to implement the code yourself? This will be better.

No, I don't. I am sure you can do this. The difficult part of the area
computation is already there - you wrote that. This is just a
reorganization. I will review it in any case.

galeo:
> I am sure you can do this.

Thanks. I'll give it a try, but maybe not too soon since I've been busy with some other things lately.

minad:
On 11/27/22 12:48, galeo wrote:
>     I am sure you can do this.
> 
> Thanks. I'll give it a try, but maybe not too soon since I've been busy
> with some other things lately.

Sure, take your time.

minad:
@galeo Thanks. I reworked it a bit and merged it.

galeo:
> I reworked it a bit and merged it.

Thanks.

> I would appreciate if you give it another thorough test.

I will.

PR #270, opened 2022-12-04
Remap beg/end-of-other-window for popupinfo

jdtsmith:
To go along with scroll-up/down.  These can be conveniently also be
used with numeric prefixes, so `M-3 M-end` goes 30% from the end, etc.


minad:
Thanks for this. I'd like to add it, but I've got warnings in package-lint.

minad:
The reason is that the command is marked as interactive-only, so we should not use it.

PR #272, opened 2022-12-05
Limit the vertical area for left/right popupinfo frame to parent frame's

### LINKED ISSUE #271: popupinfo vertical alignment when falling off frame
When setting `corfu-popup-min-height` to something large (like 15), popups near the bottom of the frame "fall off" (off the physical display too usually).  In order to keep the popup "in frame" more often, an idea:

- if the corfu height is smaller than the popupinfo's, and the min or actual height would place part of it vertically off the content frame, align the _bottom_ of the popup with the bottom of corfu's frame.  
- If, in this bottom-alignment case, a popup would also fall off the top of the content frame, top-alignment wins. 
 
This is similar to how the current placement tries to keep the child-frame on the content frame left-right.  Here a popupinfo falls off the bottom of my display:

image

This is particularly an issue for shells, with `comint-scroll-show-maximum-output` enabled, since input will often be near the bottom of the window.

jdtsmith:
This simple change alters the computed area available for `left` and `right` `popupinfo` frames, limiting the height to the remaining vertical extent of the parent frame.  In practice, this can mean `vertical` placement can be used on an emergency basis when `corfu-popupinfo-min-height` is large, even if it is not configured in `corfu-popupinfo-direction`.  Fixes #271.

### DIFF
--- a/extensions/corfu-popupinfo.el
+++ b/extensions/corfu-popupinfo.el
@@ -391,7 +391,8 @@ visible, the other window is moved to beginning or end."
   (if (corfu-popupinfo--visible-p)
       (with-selected-frame corfu-popupinfo--frame
         (with-current-buffer " *corfu-popupinfo*"
-          (with-no-warnings (end-of-buffer n))))
+          (with-no-warnings
+	    (end-of-buffer n))))
     (end-of-buffer-other-window n)))
 
 (defun corfu-popupinfo-beginning (&optional n)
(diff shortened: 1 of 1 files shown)


jdtsmith:
Here is an example of an overly-tall popup being forced vertical to avoid "falling off the screen".  Here `popupinfo-direction = (right left)`: 

image

Works well here, but could use testing with other placement width/height settings.

PR #273, opened 2022-12-05
Align popups at the bottom when the doc popup is taller than the candidates popup and limit height of the doc popup to avoid exceeding the parent frame

galeo:
See the discussion [here](https://github.com/minad/corfu/issues/271#issuecomment-1337102561).


galeo:
Added a commit to make sure y value is non-negative.

galeo:
The third commit ensures that the height of the doc popup doesn't get too large to exceed the parent frame.

minad:
Thanks

PR #275, opened 2022-12-11
Add support for asynchronous Capfs

minad:
@jdtsmith I quickly wrote a small prototype. It is completely untested. 

There is a small difference in comparison to your proposal. Instead of `:before-complete` I use a symbol property `async-completion-at-point-function` since we have to determine beforehand if a Capf is actually asynchronous such that we don't have to walk through the entire `completion-at-point-functions` list. The asynchronous Capf is then called with a callback and must return a cancellation function. 

The addition is simple enough (addition of ~10 lines), but cannot be implemented as an extension. This still leaves us with a chicken and egg problem. Will there a backend which will implement this any time soon?

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -1139,19 +1139,12 @@ ASYNC may be an asynchronous capf result."
                   (lambda ()
                     ;; Ask the backend for refreshing
                     (funcall fun (lambda (ret)
+                                   ;; TODO ensure that this runs in the correct buffer.
                                    (pcase ret
                                      (`(,beg ,end ,table . ,plist)
-                                      ;; TODO ensure that this runs in the correct buffer.
-                                      ;; TODO check validity of boundaries
-                                      ;; If valid, update completion data and UI.
-                                      (setq completion-in-region--data
-                                            (list (if (markerp beg) beg (copy-marker beg))
-                                                  (copy-marker end t)
-                                                  table
-                                                  (plist-get plist :predicate))
-                                            corfu--extra plist)
+                                      ;; TODO Check validity of boundaries.
+                                      ;; If valid, update UI. Otherwise quit.
                                       (corfu--post-command))
-                                     ;; TODO ensure correct buffer
                                      (_ (corfu-quit)))))
                     ;; Return t, since the predicate is invoked asynchronously
                     t)
--- a/corfu.el
+++ b/corfu.el
@@ -1136,7 +1136,9 @@ RESULT may be a capf result, if already present."
       (`(,fun ,beg ,end ,table . ,plist)
        (let ((completion-in-region-mode-predicate
               (or (plist-get plist :continue-predicate)
-                  (if result fun (lambda () (eq beg (car-safe (funcall fun)))))))
+                  (if fun
+                      (lambda () (eq beg (car-safe (funcall fun))))
+                    (lambda () t))))
              (completion-extra-properties plist))
          (setq completion-in-region--data
                (list (if (markerp beg) beg (copy-marker beg))
@@ -1161,7 +1163,7 @@ RESULT may be a capf result, if already present."
             (run-hook-wrapped 'completion-at-point-functions
                               #'corfu--auto-capf-wrapper-async
                               (lambda (result)
-                                (corfu--auto-deferred tick (or result t)))))
+                                (corfu--auto-deferred tick (cons nil result)))))
       (unless corfu--auto-cancel
         (if (<= corfu-auto-delay 0)
             (corfu--auto-deferred tick)
--- a/corfu.el
+++ b/corfu.el
@@ -232,8 +232,8 @@ The completion backend can override this with
     map)
   "Corfu keymap used when popup is shown.")
 
-(defvar corfu--auto-timer nil
-  "Auto completion timer.")
+(defvar corfu--auto-cancel nil
+  "Auto completion cancellation function.")
 
 (defvar-local corfu--candidates nil
   "List of candidates.")
@@ -1125,42 +1125,55 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
     (define-key map (vector last-command-event) replace)
     (funcall replace)))
 
-(defun corfu--auto-complete-deferred (&optional tick)
+(defun corfu--auto-deferred (tick)
   "Initiate auto completion if TICK did not change."
-  (setq corfu--auto-timer nil)
+  (setq corfu--auto-cancel nil)
   (when (and (not completion-in-region-mode)
-             (or (not tick) (equal tick (corfu--auto-tick))))
-    (pcase (while-no-input ;; Interruptible capf query
-             (run-hook-wrapped 'completion-at-point-functions #'corfu--capf-wrapper))
-      (`(,fun ,beg ,end ,table . ,plist)
-       (let ((completion-in-region-mode-predicate
-              (lambda () (eq beg (car-safe (funcall fun)))))
-             (completion-extra-properties plist))
-         (setq completion-in-region--data
-               (list (if (markerp beg) beg (copy-marker beg))
-                     (copy-marker end t)
-                     table
-                     (plist-get plist :predicate)))
-         (corfu--setup)
-         (corfu--exhibit 'auto))))))
+             (equal tick (corfu--auto-tick)))
+    (corfu--auto-result
+     (while-no-input ;; Interruptible capf query
+       (run-hook-wrapped 'completion-at-point-functions #'corfu--capf-wrapper)))))
+
+(defun corfu--auto-result (result)
+  "Handle capf RESULT for auto completion."
+  (pcase result
+    (`(,fun ,beg ,end ,table . ,plist)
+     (let ((completion-in-region-mode-predicate
+            (lambda () (eq beg (car-safe (funcall fun)))))
+           (completion-extra-properties plist))
+       (setq completion-in-region--data
+             (list (if (markerp beg) beg (copy-marker beg))
+                   (copy-marker end t)
+                   table
+                   (plist-get plist :predicate)))
+       (corfu--setup)
+       (corfu--exhibit 'auto)))))
 
 (defun corfu--auto-post-command ()
   "Post command hook which initiates auto completion."
-  (when corfu--auto-timer
-    (cancel-timer corfu--auto-timer)
-    (setq corfu--auto-timer nil))
+  (when corfu--auto-cancel
+    (funcall corfu--auto-cancel)
+    (setq corfu--auto-cancel nil))
   (when (and (not completion-in-region-mode)
              (not defining-kbd-macro)
              (not buffer-read-only)
              (corfu--match-symbol-p corfu-auto-commands this-command)
              (corfu--popup-support-p))
-    (if (<= corfu-auto-delay 0)
-        (corfu--auto-complete-deferred)
-      ;; NOTE: Do not use idle timer since this leads to unacceptable slowdowns,
-      ;; in particular if flyspell-mode is enabled.
-      (setq corfu--auto-timer
-            (run-at-time corfu-auto-delay nil
-                         #'corfu--auto-complete-deferred (corfu--auto-tick))))))
+    (let ((tick (corfu--auto-tick)))
+      (setq corfu--auto-cancel
+            (run-hook-wrapped 'completion-at-point-functions
+                              #'corfu--auto-capf-wrapper-async
+                              (lambda (result)
+                                (setq corfu--auto-cancel nil)
+                                (when (and (not completion-in-region-mode)
+                                           (equal tick (corfu--auto-tick)))
+                                  (corfu--auto-result result)))))
+      (unless corfu--auto-cancel
+        (if (<= corfu-auto-delay 0)
+            (corfu--auto-deferred tick)
+          (let ((timer (run-at-time corfu-auto-delay nil
+                                    #'corfu--auto-deferred tick)))
+            (setq corfu--auto-cancel (lambda () (cancel-timer timer)))))))))
 
 (defun corfu--auto-tick ()
   "Return the current tick/status of the buffer.
@@ -1185,6 +1198,11 @@ Auto completion is only performed if the tick did not change."
    (t
     (remove-hook 'post-command-hook #'corfu--auto-post-command 'local)
     (kill-local-variable 'completion-in-region-function))))
+
+(defun corfu--auto-capf-wrapper-async (fun callback)
+  "Call asynchronous capf FUN with CALLBACK."
+  (when (and (symbolp fun) (get fun 'corfu-async))
+    (funcall fun callback)))
 
 (defun corfu--capf-wrapper (fun &optional prefix)
   "Wrapper for `completion-at-point' FUN.
--- a/corfu.el
+++ b/corfu.el
@@ -1161,7 +1161,7 @@ RESULT may be a capf result, if already present."
     (let ((tick (corfu--auto-tick)))
       (setq corfu--auto-cancel
             (run-hook-wrapped 'completion-at-point-functions
-                              #'corfu--auto-capf-wrapper-async
+                              #'corfu--async-capf-wrapper
                               (lambda (result)
                                 (corfu--auto-deferred tick (cons nil result)))))
       (unless corfu--auto-cancel
@@ -1195,7 +1195,7 @@ Auto completion is only performed if the tick did not change."
     (remove-hook 'post-command-hook #'corfu--auto-post-command 'local)
     (kill-local-variable 'completion-in-region-function))))
 
-(defun corfu--auto-capf-wrapper-async (fun cb)
+(defun corfu--async-capf-wrapper (fun cb)
   "Call asynchronous capf FUN with CB."
   (and (symbolp fun) (get fun 'async-completion-at-point-function) (funcall fun cb)))
 
--- a/corfu.el
+++ b/corfu.el
@@ -1192,10 +1192,9 @@ Auto completion is only performed if the tick did not change."
     (remove-hook 'post-command-hook #'corfu--auto-post-command 'local)
     (kill-local-variable 'completion-in-region-function))))
 
-(defun corfu--auto-capf-wrapper-async (fun callback)
-  "Call asynchronous capf FUN with CALLBACK."
-  (when (and (symbolp fun) (get fun 'async-completion-at-point-function))
-    (funcall fun callback)))
+(defun corfu--auto-capf-wrapper-async (fun cb)
+  "Call asynchronous capf FUN with CB."
+  (and (symbolp fun) (get fun 'async-completion-at-point-function) (funcall fun cb)))
 
 (defun corfu--capf-wrapper (fun &optional prefix)
   "Wrapper for `completion-at-point' FUN.
--- a/corfu.el
+++ b/corfu.el
@@ -1125,29 +1125,25 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
     (define-key map (vector last-command-event) replace)
     (funcall replace)))
 
-(defun corfu--auto-deferred (tick)
-  "Initiate auto completion if TICK did not change."
+(defun corfu--auto-deferred (tick &optional result)
+  "Initiate auto completion if TICK did not change.
+RESULT may be a capf result, if already present."
   (setq corfu--auto-cancel nil)
-  (when (and (not completion-in-region-mode)
-             (equal tick (corfu--auto-tick)))
-    (corfu--auto-result
-     (while-no-input ;; Interruptible capf query
-       (run-hook-wrapped 'completion-at-point-functions #'corfu--capf-wrapper)))))
-
-(defun corfu--auto-result (result)
-  "Handle capf RESULT for auto completion."
-  (pcase result
-    (`(,fun ,beg ,end ,table . ,plist)
-     (let ((completion-in-region-mode-predicate
-            (lambda () (eq beg (car-safe (funcall fun)))))
-           (completion-extra-properties plist))
-       (setq completion-in-region--data
-             (list (if (markerp beg) beg (copy-marker beg))
-                   (copy-marker end t)
-                   table
-                   (plist-get plist :predicate)))
-       (corfu--setup)
-       (corfu--exhibit 'auto)))))
+  (when (and (not completion-in-region-mode) (equal tick (corfu--auto-tick)))
+    (pcase (or result (while-no-input ;; Interruptible capf query
+                        (run-hook-wrapped 'completion-at-point-functions
+                                          #'corfu--capf-wrapper)))
+      (`(,fun ,beg ,end ,table . ,plist)
+       (let ((completion-in-region-mode-predicate
+              (lambda () (eq beg (car-safe (funcall fun)))))
+             (completion-extra-properties plist))
+         (setq completion-in-region--data
+               (list (if (markerp beg) beg (copy-marker beg))
+                     (copy-marker end t)
+                     table
+                     (plist-get plist :predicate)))
+         (corfu--setup)
+         (corfu--exhibit 'auto))))))
 
 (defun corfu--auto-post-command ()
   "Post command hook which initiates auto completion."
@@ -1164,10 +1160,7 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
             (run-hook-wrapped 'completion-at-point-functions
                               #'corfu--auto-capf-wrapper-async
                               (lambda (result)
-                                (setq corfu--auto-cancel nil)
-                                (when (and (not completion-in-region-mode)
-                                           (equal tick (corfu--auto-tick)))
-                                  (corfu--auto-result result)))))
+                                (corfu--auto-deferred tick (or result t)))))
       (unless corfu--auto-cancel
         (if (<= corfu-auto-delay 0)
             (corfu--auto-deferred tick)
--- a/corfu.el
+++ b/corfu.el
@@ -1125,20 +1125,37 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
     (define-key map (vector last-command-event) replace)
     (funcall replace)))
 
-(defun corfu--auto-deferred (tick &optional result)
+(defun corfu--auto-deferred (tick &optional async)
   "Initiate auto completion if TICK did not change.
-RESULT may be a capf result, if already present."
+ASYNC may be an asynchronous capf result."
   (setq corfu--auto-cancel nil)
   (when (and (not completion-in-region-mode) (equal tick (corfu--auto-tick)))
-    (pcase (or result (while-no-input ;; Interruptible capf query
+    (pcase (or async (while-no-input ;; Interruptible capf query
                         (run-hook-wrapped 'completion-at-point-functions
                                           #'corfu--capf-wrapper)))
       (`(,fun ,beg ,end ,table . ,plist)
        (let ((completion-in-region-mode-predicate
-              (or (plist-get plist :continue-predicate)
-                  (if fun
-                      (lambda () (eq beg (car-safe (funcall fun))))
-                    (lambda () t))))
+              (if async
+                  (lambda ()
+                    ;; Ask the backend for refreshing
+                    (funcall fun (lambda (ret)
+                                   (pcase ret
+                                     (`(,beg ,end ,table . ,plist)
+                                      ;; TODO ensure that this runs in the correct buffer.
+                                      ;; TODO check validity of boundaries
+                                      ;; If valid, update completion data and UI.
+                                      (setq completion-in-region--data
+                                            (list (if (markerp beg) beg (copy-marker beg))
+                                                  (copy-marker end t)
+                                                  table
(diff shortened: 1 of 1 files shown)


minad:
_Originally posted by @jdtsmith in https://github.com/minad/corfu/issues/103#issuecomment-1345423053_

## Wish

A very simple and limited form of async support facilitating _cache pre-warming_ of completions.  

## Current situation

Right now corfu-auto sets up a timer, and, when the timer ticks over, consults the CAPFs.  If a CAPF says it can complete there, corfu requests the completions, and then pops them up.  But there is a problem when the CAPF table is gathering completions from an external process — it usually doesn't yet _know_ the completions.  But corfu urgently needs them.  The only choice the CAPF's table function has is to BLOCK until the process gives it what it needs.  There is not option to tell corfu: "I'll get back to you".

For many situations the block is unnoticeable (ms or 10s of ms), but sometimes (depending on process vagaries), it could be much more (2s sometimes in my usage).  To help, CAPF authors can do 2 things:

1. Make this blocking process call interruptible (inhibit-quit=nil),
2. Use `while-no-input` to abort on input.  

At least in this way the user is not frozen indefinitely.

## Problem

`while-no-input` has at fastest a ~1sec response on my system.   And after you type another character in... well the process starts all over again.  Typing with 1sec between keystrokes feels awful.  So high latency CAPFs get tossed overboard.

## Simple idea

A mode can provide to corfu an optional function via a special CAPF return _property_: `:before-complete` [1].  

If such a function is configured in the returned CAPF results, the process goes like:

1. corfu's timer function calls the `:before-complete` function with a callback function (which can capture the tick data via a closure var). 
5. On the CAPF side, after it gets the completion data it needs, the CAPF calls the callback function corfu supplied it, with `t` for success or `nil` for failure (optional: it can also just not call-back for failure).
6. In the callback itself, if the `:before-complete` function succeeded (and the tick data match), it goes on to call `corfu--auto-complete-deferred`, popping up completions, just as normal (if a bit delayed).
7. This _still_ checks the buffer tick to make sure nothing has changed, just like before.  

An analogous route for "on-demand" completion (`M-Tab`) completion would also be straightforward (though perhaps not necessary and unintuitive if it never pops up).

## Why?

With this simple support, a CAPF that routinely has high latency can configure a `:before-complete` function which pre-warms its cache and then calls back to corfu to continue, once it has (presumably asynchronously) gathered the relevant information. 

### Outcomes 

**What this does not do**: it does not decrease the actual latency before completions pop-up, for high latency processes. 

**What this does do**: it permits CAPF designs which do not block user input at all, as completion data are awaited fully asynchronously, if the CAPF is designed with latency in mind.

**What about other UIs**: They will still work.  They will ignore `:before-complete` and just directly call the CAPF table function, which will block (and cause people to say: just use corfu).

**Effects users may notice**: It may sometimes take longer for corfu to popup than usual.

**Complexities**: Suppose the timer ticks over and a CAPF's `:before-complete` function has not yet called the callback.  If buffer-tick is unchanged from the one sent to `:before-complete` there is nothing to do, just keep waiting. If the tick _has_ changed, well the old attempt can safely be ignore, and a new completion started.  A good CAPF will notice a new request to `:before-complete` coming in and cancel any that are pending, but that's not essential.

[1] could prefix it like `:corfu-before-complete` if you wanted to parallel company.

minad:
A simple test capf:

~~~ elisp
(defun async-test-capf (&optional callback)
  (if callback
      (let ((timer (run-at-time
                    0.01 nil
                    (lambda ()
                      (funcall callback (cons #'always (cape-symbol)))))))
        (lambda () (cancel-timer timer)))
    (cape-symbol)))

(put #'async-test-capf 'async-completion-at-point-function t)

(setq-local completion-at-point-functions (list #'async-test-capf))
~~~

minad:
>  1. I'm not clear why you need to pre-filter async CAPFs. In my
>     conception, if the "winning" CAPF returns a |:before-complete| (or
>     whatever it's called) property, then corfu can hand the specified
>     function the callback and quickly return, letting it followup via
>     the callback. If there is no such property, corfu can just proceed
>     directly with completion the "old-fashioned" way.

The problem is that going through the list of Capfs is potentially slow.
Not every Capf tries to be fast, some already perform prefiltering.
Therefore I separated this completely. Note that the asynchronous Capfs
are already called directly in the post-command-hook. It is of crucial
importance that there is no slowness in the post-command-hook at all. In
contrast, in the current implementation we call the Capfs only after the
timer elapsed.

>  2. In my original idea, the CAPF itself doesn't get /any/ async
>     support, only its /table function/ does. After all /"The functions
>     on this hook should generally return quickly, since they may be
>     called very often"/. So corfu can freely call the CAPF whenever it
>     wants to check for "still valid" or not, and it will be expected to
>     return something immediately.

Not true unfortunately. We could however rely on this requirement for
the correct function of async capfs. But I would rather not do that
since it will lead to a more brittle system.

> In my view this means the CAPF itself is not (or should not be) the one
> in need of async help. In my situation, my CAPF would be happy to report
> whether it is still applicable (not in a string or comment, say), even
> though its /table/ would benefit from async support for pre-computing
> completions.

The design I tried to implement here somehow resembles the design of
lsp-bridge or it should at least allow to glue this together. Assume
that when the Capf is called, the backend is asked for possible
completions at point. As soon as the server returns a static table is
passed to the callback. In the next step we could then try to make the
table itself asynchronous, but this will get much more complex due to
all the metadata functions and table actions. I attempted this when we
discussed the cape-async wrapper.

> The only addition I was conceiving to my mode is to call my normal
> pre-existing cache-filling code that I already use, but "earlier" (i.e.
> before the completion UI pops up) and without blocking. If the cache
> misses, e.g. because it was never computed, I'll still block and wait
> for results, just like before. But if the cache hits, I can immediately
> return the results. Basically, the idea was to magically reduce typing
> latency in a way that's compatible with company and other completion
> systems.

I think it should also be possible to adapt your backend to the
alternative design I am considering here. I think in the end both design
are somewhat equivalent, but my design seems a bit cleaner/easier to use.

Your design: The prefetcher (:before-complete) is called before
completion. The capf should return nil (or even block) as long as no
data is available. As soon as data becomes available, the callback
notifies the UI.

My design: The asynchronous capf called with a callback argument,
returns a result via the callback as soon it becomes available. The
backend becomes a lot simpler since we don't have to bother with this
blocking logic. However this advantage is not entirely free since we
still have to implement the blocking code path anyway for the
synchronous capf. My design just untangles the two different code paths.
Turning a purely asynchronous capf into a synchronous+asynchronous one
should be easy:

(defun make-synchronous (async-capf)
  (lambda (&interactive callback)
    (if callback
        (funcall async-capf callback)
      (let ((result 'wait))
        (funcall async-capf (lambda (ret) (setq result ret)))
        (while (eq result 'wait)
          (sit-for 0.01))
        result))))

> Note this is a softer/easier form of |.then| style async compared to a
> system that can update results in "mid-flight". But I really believe it
> will help with many cases. Something like lsp-bridge will improve the
> /absolute/ latency, but this should mostly eliminate "I'm typing through
> molasses", which is what users (myself included) especially hate. I
> /think/ this is pretty similar to company backend's async support
> <https://www.badykov.com/emacs/async-company-mode-backend/>.

The Company backend is similar to fully asynchronous tables where every
operation is made asynchronous (cape-async-capf). However the issue with
this design is that most of the operations are not actually called in an
asynchronous fashion. Instead most of them must be forcibly
synchronized, e.g., annotation computation or the boundary checks.

In contrast, the design I've implemented in this PR introduces
controlled asynchronicity at a single point (the entry point), where we
can guarantee that the UI takes advantage of it.

All the remaining operations remain synchronous. This is not an issue as
long as all the data is already be available as soon as the backend
process returned them. Here I assume that the post filtering is then
happening only on the already returned data (Orderless style). In order
to support updating candidates asynchronously/dynamically one could
allow the callback to be called again. In this prototype the callback is
single-shot, but we could as well generalize this. The first call to the
callback returns a table as soon as completion becomes possible,
subsequent calls update the table and force a refresh.

jdtsmith:
> Note that the asynchronous Capfs are already called directly in the post-command-hook.

Is that because you want to give CAPFs a "head start" to pre-compute things so they are more likely to provide results "on time"?  The problem I see with this is you can't then use `corfu-auto-delay` to control how frequently your process is getting pestered.  I was thinking `corfu-auto-delay` would then become the _minimum_ delay while idle for completions to popup, and to recommend to users not to set this too low.  Trying to compute completions data on every change in the PCH will be too heavy I fear (I have to send a keyboard interrupt to get out of an ongoing completion query, for example). 

I'll put a few more comments in the code.

minad:
> Is that because you want to give CAPFs a "head start" to pre-compute things so they are more likely to provide results "on time"? The problem I see with this is you can't then use corfu-auto-delay to control how frequently your process is getting pestered. I was thinking corfu-auto-delay would then become the minimum delay while idle for completions to popup, and to recommend to users not to set this too low. Trying to compute completions data on every change in the PCH will be too heavy I fear (I have to send a keyboard interrupt to get out of an ongoing completion query, for example).

There won't be such a thing as corfu-auto-delay for asynchronous capfs. The capfs will have full control.

I had another idea how the refreshing could work. We also turn the predicate asynchronous, see https://github.com/minad/corfu/pull/275/commits/a501b92ee33227ca720da389e69991863498d0c2. Now there is only a single extension, the `async-completion-at-point-function` property. The UI recalls the capf after every change and the capf should then pass an updated table or nil to the callback.

jdtsmith:
> Your design: The prefetcher (:before-complete) is called before
completion. The capf should return nil (or even block) as long as no
data is available. As soon as data becomes available, the callback
notifies the UI.

Small correction: in my design, the capf returns limits, a table function, and properties if it can (eventually) complete there, as normal; so for non-async UI's, it's logic is unchanged.  If an only if :before-complete exists, the callback it is provided later notifies the UI, which goes ahead to activate completion-in-region mode, and call the table (essentially the `corfu--setup`/`corfu--exhibit` stuff).

corfu.el
```diff
@@ -1125,16 +1125,31 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
...
+                        (run-hook-wrapped 'completion-at-point-functions
+                                          #'corfu--capf-wrapper)))
       (`(,fun ,beg ,end ,table . ,plist)
        (let ((completion-in-region-mode-predicate
-              (lambda () (eq beg (car-safe (funcall fun)))))
+              (if async
+                  (lambda ()
+                    ;; TODO Implement cancellation.
+                    ;; Ask the backend for refreshing
+                    (funcall fun (lambda (ret)
```
jdtsmith, line 1142: How is an ASYNC capf to know whether it is being called simply for the side effect of knowing if completion is still valid here, vs. when the UI intends to ask for completion data (after the async CAPF responds, via the callback)?  One simple idea is to pass no callback here, i.e. call the CAPF synchronously.  

I suppose it's possible a CAPF doesn't yet know if it can complete there, and has to ask its process.  Honestly that's my actual situation, but I have some elaborate "guess then correct with boundaries" approach that works... maybe it could be abandoned.

minad:
On 12/11/22 21:45, JD Smith wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In corfu.el
> <https://github.com/minad/corfu/pull/275#discussion_r1045294101>:
> 
>>        (`(,fun ,beg ,end ,table . ,plist)
>         (let ((completion-in-region-mode-predicate
> -              (lambda () (eq beg (car-safe (funcall fun)))))
> +              (if async
> +                  (lambda ()
> +                    ;; TODO Implement cancellation.
> +                    ;; Ask the backend for refreshing
> +                    (funcall fun (lambda (ret)
> 
> How is an ASYNC capf to know whether it is being called simply for the
> side effect of knowing if completion is still valid here, vs. when the
> UI intends to ask for completion data (after the async CAPF responds,
> via the callback)? One simple idea is to pass no callback here, i.e.
> call the CAPF synchronously.
> 
> I suppose it's possible a CAPF doesn't yet know if it can complete
> there, and has to ask its process. Honestly that's my actual situation,
> but I have some elaborate "guess then correct with boundaries" approach
> that works... maybe it could be abandoned.

Well, the situation is just the same for the synchronous capf. I think
it is actually easy. The capf has to maintain some global state, which
is updated on each capf call. In the synchronous case, it should block
and return the current table (or nil). In the asynchronous case, it
should return return immediately with a cancellation function and then
pass the current table (or nil) to the callback as soon as the result is
available.

However one could argue that the synchronous capfs are more often static
in nature and pure, and don't have to maintain this global state. The
asynchronous capf in contrast has to maintain state such that the table
is updated accordingly. In order to reflect this in the API design we
could also add a :async-predicate function returned from the initial
capf call, which could then access the local state (closure) constructed
during the initial capf call.

corfu.el
```diff
@@ -1146,21 +1161,27 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
...
-      (setq corfu--auto-timer
-            (run-at-time corfu-auto-delay nil
-                         #'corfu--auto-complete-deferred (corfu--auto-tick))))))
+    (let ((tick (corfu--auto-tick)))
+      (setq corfu--auto-cancel
+            (run-hook-wrapped 'completion-at-point-functions
+                              #'corfu--async-capf-wrapper
+                              (lambda (result)
+                                ;; TODO ensure that this runs in the correct buffer
+                                (corfu--auto-deferred tick (or result t)))))
```
jdtsmith, line 1178: I am not sure about calling the async-only CAPFs on every keypress.  That's way too expensive for launching one-at-a-time commands over a shell (my setup).  I suppose I can keep my own timer going and simply return `nil` if a new call comes too soon (and you come asking with a CB supplied).  My CAPF can also I suppose decide that the new call is a subset of the current call, so simply "do nothing", i.e. return nil to let corfu wait for the original call to complete.  But I don't think that's an option here: that will "fall back" on the regular auto code below, which will just call the CAPF synchronously every corfu-auto-delay sec.  Clearly that's not desired.

corfu.el
jdtsmith: You don't want all the synchronous CAPFs on the list to "pay the price" of waiting for one async CAPF to find out whether it can complete.  I suppose a good and kind CAPF would do some simple checks first (in a string?  in a comment?) and return a quick nil.  But (see below), returning nil for an ASYNC CAPF doesn't do what you want. This is the reason I thought that the original capf call should be sync, with an async "pause to gather results" before the table is consulted.

minad:
On 12/11/22 22:04, JD Smith wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In corfu.el
> <https://github.com/minad/corfu/pull/275#discussion_r1045296289>:
> 
>>        (`(,fun ,beg ,end ,table . ,plist)
>         (let ((completion-in-region-mode-predicate
> -              (lambda () (eq beg (car-safe (funcall fun)))))
> +              (if async
> +                  (lambda ()
> +                    ;; TODO Implement cancellation.
> +                    ;; Ask the backend for refreshing
> +                    (funcall fun (lambda (ret)
> 
> You don't want all the synchronous CAPFs on the list to "pay the price"
> of waiting for one async CAPF to find out whether it can complete. I
> suppose a good and kind CAPF would do some simple checks first (in a
> string? in a comment?) and return a quick nil. But (see below),
> returning nil for an ASYNC CAPF doesn't do what you want. This is the
> reason I thought that the original capf call should be sync, with an
> async "pause to gather results" before the table is consulted.

The idea would be that an async capf is always exclusive. There is no
alternative to that. However one could then use a merging mechanism like
cape-super-capf.

minad:
On 12/11/22 21:55, JD Smith wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In corfu.el
> <https://github.com/minad/corfu/pull/275#discussion_r1045295327>:
> 
>>    (when (and (not completion-in-region-mode)
>               (not defining-kbd-macro)
>               (not buffer-read-only)
>               (corfu--match-symbol-p corfu-auto-commands this-command)
>               (corfu--popup-support-p))
> -    (if (<= corfu-auto-delay 0)
> -        (corfu--auto-complete-deferred)
> -      ;; NOTE: Do not use idle timer since this leads to unacceptable slowdowns,
> -      ;; in particular if flyspell-mode is enabled.
> -      (setq corfu--auto-timer
> -            (run-at-time corfu-auto-delay nil
> -                         #'corfu--auto-complete-deferred (corfu--auto-tick))))))
> +    (let ((tick (corfu--auto-tick)))
> +      (setq corfu--auto-cancel
> +            (run-hook-wrapped 'completion-at-point-functions
> +                              #'corfu--async-capf-wrapper
> +                              (lambda (result)
> +                                ;; TODO ensure that this runs in the correct buffer
> +                                (corfu--auto-deferred tick (or result t)))))
> 
> I am not sure about calling the async-only CAPFs on every keypress.
> That's way too expensive for launching one-at-a-time commands over a
> shell (my setup). I suppose I can keep my own timer going and simply
> return |nil| if a new call comes too soon (and you come asking with a CB
> supplied). My CAPF can also I suppose decide that the new call is a
> subset of the current call, so simply "do nothing", i.e. return nil to
> let corfu wait for the original call to complete. But I don't think
> that's an option here: that will "fall back" on the regular auto code
> below, which will just call the CAPF synchronously every
> corfu-auto-delay sec. Clearly that's not desired.

There shouldn't be a problem here. The backend must be informed that the
buffer content changed and has to act accordingly. It could either
ignore the new call to the backend or it could cancel the old and still
running call to the backend. Note that packages like fish-completion
already work like this. The only difference would be that we return
asynchronously.

jdtsmith:
> My design: The asynchronous capf called with a callback argument,
returns a result via the callback as soon it becomes available. The
backend becomes a lot simpler since we don't have to bother with this
blocking logic. However this advantage is not entirely free since we
still have to implement the blocking code path anyway for the
synchronous capf. My design just untangles the two different code paths.

This seems like a fairly significant addition to CAPF semantics.  If corfu fades away (perish the thought) or a another capf-centric UI gains traction, or even just to support company-with-capf at present, it seems like preserving as much as possible of the calling semantics of CAPFs is desirable.  Also this is easier for authors.  

Hence the thought to use properties to provide simple async target points.  As long as the CAPF calls themselves remain SYNC, this is all compatible with the fact that some CAPFs are badly behaved: while-no-input can cancel them.  I guess it's not _such_ a big change to provision a capf with an optional extra `cb` argument and check for it.  But it makes it hard to "mix" sync and async on one completion-at-point-functions list.  

Example: suppose I have shell-completions (async, expensive) and file-completions (sync, cheap) configured in `completion-at-point-functions`.   With the current async design:

- on each new char press in a PCH, the async CAPF will be called and the returned cancel function saved (can it opt out?  Not apparently).  They may call back via `cb` to corfu after a second or two to continue on.
- corfu-auto, every 2 sec, will call ALL the CAPFS (even the ASYNC ones), synchronously (no callback argument). 

What is an async CAPF supposed to do when called synchronously?  Lie and say it can't complete there (even though it could)?  It doesn't want to block, but it will have to.  What if COMPANY is the one calling?  What if the user has just used `M-Tab`?   Blocking is the only choice.

All of this, including mixing sync and async in one list, tends IMO to lead one down the simple road of:

1. SYNC CAPF call,
2. ASYNC UI PAUSE,
3. SYNC TABLE calls (hopefully now with cache in hand)

Yes this requires a CAPF to know "right now" that it can in principle complete (or not) here.  But in practice I think that's usually simpler than computing actual completions. 

Maybe we should make a toy ASYNC caching capf to try hooking up in various ways.

minad:
On 12/11/22 22:24, JD Smith wrote:
> This seems like a fairly significant addition to CAPF semantics. If
> corfu fades away (perish the thought) or a another capf-centric UI gains
> traction, or even just to support company-with-capf at present, it seems
> like preserving as much as possible of the calling semantics of CAPFs is
> desirable. Also this is easier for authors.

I argue that making the capf function itself asynchronous is the
simplest addition we could do which gives us fully asynchronous
completion. The approach with properties won't work as general (needs
blocking etc) and also isn't as efficient since we still have to call
all the capfs.

> Hence the thought to use properties to provide simple async target
> points. As long as the CAPF calls themselves remain SYNC, this is all
> compatible with the fact that some CAPFs are badly behaved:
> while-no-input can cancel them. I guess it's not /such/ a big change to
> provision a capf with an optional extra |cb| argument and check for it.
> But it makes it hard to "mix" sync and async on one
> completion-at-point-functions list.

That's certainly a problem. But it should be possible to implement this
such that the capfs are tried in order. The synchronous capfs coming
after an asynchronous one have to wait until the asynchronous one
returned with a nil result.

What we would do this basically to turn the run-hook-wrapped into
continuation passing style. Synchronous capfs immediately call the
return continuation, asynchronous capfs call the continuation as soon as
the result becomes available. Since we only do a transformation from
direct style to cps is also why I believe that this is the simplest
possible change we could do. Also it would be an elegant change, not an
adhoc addition.

> What is an async CAPF supposed to do when called synchronously? Lie and
> say it can't complete there (even though it could)? It doesn't want to
> block, but it will have to. What if COMPANY is the one calling? What if
> the user has just used |M-Tab|? Blocking is the only choice.

Yes, async capfs called synchronously must block to preserve the
existing API. But that's just as good as what we have right now.

> All of this, including mixing sync and async in one list, tends IMO to
> lead one down the simple road of:
> 
>  1. SYNC CAPF call,
>  2. ASYNC UI PAUSE,
>  3. SYNC TABLE calls (hopefully now with cache in hand)
> 
> Yes this requires a CAPF to know "right now" that it can in principle
> complete (or not) here. But in practice I think that's usually simpler
> than computing actual completions.

It is an adhoc addition and it is not general, since it needs this
guessing. I understand that it would fix your particular backend. But
given that we could maybe achieve a more general solution I don't think
we should add a :before-complete extension.

jdtsmith:
> Well, the situation is just the same for the synchronous capf. 

I think that's not only true if synchronous CAPFs don't return quickly on Step 1 (even if Step 2 is hard for them).  So slow/misbehaving synchronous CAPFs can make other CAPFs wait.  ASYNC CAPFs in the current design will entirely block-out corfu-auto and sync CAPFs I think (unless I'm mis-reading; not tested here either).

minad:
To summarize the whole picture:

1. Either we add ad-hoc hacks like in your proposal, which may improve certain things.
2. We try to make the completion table itself asynchronous. I attempted this in Cape - we had this long discussion there on the tracker. I don't believe this is the way to go since the API is too complex, making everything asynchronous is neither feasible nor beneficial given that the UI cannot take advantage of asynchronicity for every table call. There will also be a large difficulty due to conversion-
3. We try to make the capf and only the capf asynchronous, while preserving the completion table api. I see one difficulty in how we handle the completion-in-region-mode-predicate (if completion should be continued). We probably would also have to make the predicate asynchronous.
4. One throws away the entire capf mechanism. But then we will end up with an asynchronous design anyway which will resemble 3., but probably only without the synchronous completion tables.

jdtsmith:
All good thoughts. 

First, I agree "one round of async" is enough.  If it's heavy enough to need async, backends should gather all the necessary information (kind, sig, etc.) in advance.  The only hesitation on this is that the LSP server protocol supports, for efficiency (ha),  ["resolving" completions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem_resolve) one at a time, to get more information.  Many actively do this. 

I like the CPS idea.  It's not too different than a callback design, in that the called function _can_ return without (yet) having called the continuation.  But it can also call it "right away" if it knows what the situation is.  For example, what I'd do is check `ppss` etc. to see if I'm in a comment or string, and if so, immediately call the continuation-func with `nil` (or return `nil` if no cont-func is present).

I guess corfu (or any UI) can detect that a given CAPF supports CPS/ASYNC style with the symbol property.  Another thing:

> Yes, async capfs called synchronously must block to preserve the existing API. But that's just as good as what we have right now.

Well, since we call the CAPF for two distinct reasons — 1. "test if this is a suitable completion spot" and 2. "prepare to generate completion data for me" — we have so far used the two-step to differentiate these.  For example, this is why I now generate the expensive completions in Step 2. Since the latter can be expensive and the former is supposed to be cheap, I think we'll still need some way for an ASYNC/CPS CAPF to know it is being called _for its Step 1 results only_.  That might be harder in the ASYNC CAPF "do the work up front" paradigm.  

Note that the following policy replicates only _half_ of what we'd need for Step 1-only calls: "If you know the answer is no, immediately call the c-func with nil".  That's fine, but what about when you _can_ complete, but should not bother caching completions (during a corfu popup, for example)?  We could fix corfu to stop doing that, but others may also use CAPFs in this way: get the table etc., but just ignore it.  

**IDEA**: maybe we could pass the new argument as a cons `(continuation-function . 'test-only)` (or similar) to indicate it you should just call the continuation-function (now or later) with the results, and not _bother_ pre-caching data.  Basically a guarantee that you'll not be calling the completion table.  Obviously this particular continuation function would be distinct from the "true completions" one.

Most CAPFs will always need to serve two master: async (corfu-auto style), and sync (corfu's M-Tab, other UI's).  I don't think the CPS style will be too hard to shim in, except for this last issue. It would be awkward if you had to spend a while deciding whether to corfu-quit, during which time other stuff can happen.

jdtsmith:
One other thing: I think will be important to keep the package changes required to support corfu's-CPS/ASYNC as minimal as possible.  

An (optional) argument to CAPF is OK, but I'm worried about no timer support.  Company and other pop-ups don't expose their timing details.  They only call into the CAPF when you are idle somewhere for that amount of time.  That's an important tunable to avoid "too frequent" calls.  

Starting/stopping completion, even async, will not be free.  Obviously the package implementing the CAPF could make its own timer and just say `nil` until it wants to complete, or even answer `nil` until the current completion is over, but that definitely breaks the "least changes needed" principle above.  If packages _really_ want constant calls after every change, they can set `corfu-auto-delay` to 0.  And if they want to know the state of the buffer at all times, they can set their own PCH.

jdtsmith:
Can you comment on your current async thoughts a week on?  If it's relatively simple to implement and easy to target, it could be a real distinguishing point for corfu.

minad:
I don't have plans right now. If you want to take a stab at it, feel free to create some prototypes. But to be clear, I am not interested in hacks working around Capf deficiencies , e.g. your :before-complete proposal. I think we should only add a design which somehow unifies asynchronous backends with Capfs. See also the recent lsp-bridge or [mono-complete](https://old.reddit.com/r/emacs/comments/zm2725/rfc_suggestions_while_typing/) attempts. I think none of all the current approaches get it right. lsp-bridge and mono-complete both completely reinvent the wheel, the company async design is flawed and there is a mismatch between capf and the company api, corfu is strongly focused on capf, which may be too limiting.

jdtsmith:
Agreed.  If it is to become an "accepted and documented part of the CAPF standard" it should probably involve emacs-devel as well.  Hopefully wouldn't devolve into bike-shedding.  

I do agree :before-complete is a bit of an "add-on" designed to solve one particular async-flavored problem.  But in terms of "hackishness", it strikes me as no worse than company-docsig or other such property-based CAPF protocol "add-ons".  It does have the advantage of involving very little change.  But I entirely agree, it's not what you'd start with if building an async system from scratch.

It's main limitation IMO is it solves just the one type of issue.  It occurs to me other CAPFs may have different needs than "pre-warm the cache just before table calls".  For some of them, type/kind lookup may be most expensive.  Then again, pre-warming may be the dominant need.  

If Emacs evolves a real async/await coupled nicely to process output, likely a different approach would emerge. In the meantime I will try to improve my hackish interrupt-on-input, by adding random `sit-for`'s... :).

minad:
> If it is to become an "accepted and documented part of the CAPF standard" it should probably involve emacs-devel as well.

It would be better to come up at least with a working prototype before starting any discussion.

> But in terms of "hackishness", it strikes me as no worse than company-docsig or other such property-based CAPF protocol "add-ons".

I disagree. Addons like the annotation function or company docsig etc make much more sense. They fit into the existing design. Your :before-complete does not.

> If Emacs evolves a real async/await coupled nicely to process output, likely a different approach would emerge.

Hmm, Emacs already has async/await. You can easily write a macro which transforms direct style to callback style. The main problem we are facing is finding a design which fits together with the existing infrastructure and generalizes it nicely.

jdtsmith:
I noticed this interesting built-in _async-or-not_ interface to eldoc, which sounds quite a bit like what we were discussing, with the minor twist of returning something special (here a "non-nil, non-string value") to indicate to the CAPF caller: "I am looking into this, will get back to you via callback".  

```
Each hook function is called with at least one argument CALLBACK,
a function, and decides whether to display a short doc string
about the context around point.

- If that decision can be taken quickly, the hook function may
  call CALLBACK immediately, following the protocol described
  below.  Alternatively, it may ignore CALLBACK entirely and
  return either the doc string, or nil if there’s no doc
  appropriate for the context.

- If the computation of said doc string (or the decision whether
  there is one at all) is expensive or can’t be performed
  directly, the hook function should return a non-nil, non-string
  value and arrange for CALLBACK to be called at a later time,
  using asynchronous processes or other asynchronous mechanisms.
```

With this style, compatibility with all existing CAPFs could be preserved; they could  just return whatever they normally do.  An issue is that CAPFs are currently defined with no arguments; your symbol property trick might smooth this over.

PR #276, opened 2022-12-12
README: elaborate on disabling Corfu in read-passwd minibuffers

mekeor:
I think it's bad that the code snippets in the Readme enable Corfu in read-passwd minibuffers. Let's at least elaborate on it a little.


minad:
Thanks. Please adjust the existing snippet and the description such that we end up with a coherent paragraph. There is also no need to link to an issue.

PR #331, opened 2023-05-19
README: Add reference to corfu-candidate-overlay package

adamkruszewski:
Shameless plug, hope you don't mind.

Added a link and a short description to my corfu-candidate-overlay package that provides "as you type" overlay showing the first auto-suggestion candidate from Corfu. I do not use the auto popup and wanted to have some indication that I can auto-complete *something* and whether there are more than one suggestions present. Package requires corfu-auto to be disabled, so probably won't be of wide use but still could be of interest to someone ;-)

Sneak-peak of the overlay in action:
image

Many thanks,
Adam.

### DIFF
--- a/README.org
+++ b/README.org
@@ -520,6 +520,10 @@ enhance your setup.
 - [[https://github.com/minad/vertico][Vertico]]: You may also want to look into my Vertico package. Vertico is the
   minibuffer completion counterpart of Corfu.
 
+- [[https://code.bsdgeek.org/adam/corfu-candidate-overlay][corfu-candidate-overlay]]: Shows as-you-type auto-suggestion candidate overlay
+  with a visual indication of whether there are many or exactly one candidate
+  available (works only with corfu-auto popup disabled).
+
 * Alternatives
 
 - [[https://github.com/company-mode/company-mode][Company]]: Company is a widely used and mature completion package, which


minad:
Hi Adam,

thanks for letting me know! I explored this concept in https://github.com/minad/corfu/pull/232. I didn't pursue it further for two reasons:

1. The suffix overlay doesn't play nicely with non-prefix completion, e.g., Orderless, Substring or Flex. On the other hand, prefix completions are common when triggering completion.

2. I didn't manage to achieve acceptable performance. The problem is that the overlay is directly shown behind the input, which requires that the overlay is updated immediately. Otherwise the input and overlay will be inconsistent which does not look right. In contrast, the popup below or above the input can lag a little bit behind (and the computation can be interruptible).

I wonder if you are negatively affected by these two issues. I am happy to link to your package, as soon as it is published on a package archive. Do you plan to add it to MELPA, GNU ELPA or NonGNU ELPA?

minad:
> I'm not myself relying on non-prefix completion so not explored the concept. 

My solution was to just show the entire candidate behind the input, e.g., if you type `flx` it will display `flx<flexible>`. Unfortunately this does not play at all together with the input faking. All we can do is push the `<flexible>` overlay to the right if there is more import, until we get an updated candidate.

> The inconsistency will always be there if you type fast and allow the backends to be interruptible, which is necessary for the performance - what I do is cache the last full suggestion and basically 'fake' that the letters typed are 'taken' out of the suggestion (or 'given' back when deleting characters) -- at least it looks OK, even if inconsistent. 

Okay, I had tried something similar. So you always override the overlay character even if it doesn't match the input character? For example if you have `pre<fix>` and type `f` it will be consistently updated to `pref<ix>`, which makes sense. But if you type `p` it will be updated inconsistently to `prep<ix>`? I didn't like the idea of doing such an inconsistent update.

> I'm using Emacs built from the master (got some optimizations recently in the department of overlays) with native compilation and have only few completion backends active...

The overlay optimizations should not be relevant. They only help if you have many overlays since they improve the scaling. Native compilation and using only few well-behaved backends certainly helps.

> I'll have to read on the MELPA or GNU ELPA process -- will do it over the weekend and pursue it.

Thanks, please keep me updated.

adamkruszewski:
> Okay, I had tried something similar. So you always override the overlay character even if it doesn't match the input character? For example if you have pre<fix> and type f it will be consistently updated to pref<ix>, which makes sense. But if you type p it will be updated inconsistently to prep<ix>? I didn't like the idea of doing such an inconsistent update.

Yes, it works exactly like this.

Probably the only alternative is to hide the overlay - tried it but causes flickering and in my opinion is more distracting than the inconsistent state. I may try to look if we would substitute the inconsistent state with a static placeholder like "..." - doubt it will look better.

minad:
Any progress here? Please ping me again when your package is ready, then I will add a link. Thanks!

adamkruszewski:
Sorry, had a hectic week at "day-job" -- should be able to pick this up near or at weekend. Thanks!

adamkruszewski:
Sorry it took so long - the PR into MELPA was just merged: https://github.com/melpa/melpa/pull/8613

Thanks for pointing me there -- it helped quite a lot in terms of quality of the package to go through the process and I have got a great suggestion/feedback already as well 😸

adamkruszewski:
Thank you. Have a wonderful day!

minad:
Thanks! I don't understand the following paragraph in your readme though.

> When using slow or cpu-heavy corfu backends (e.g. [cape's](https://github.com/minad/cape) dict backend with a larger dictionary file[4](https://code.bsdgeek.org/adam/corfu-candidate-overlay#user-content-footnote-4)) you may experience lagging when writing fast contrary to normal experience with standard corfu with auto-suggestions popup enabled. This is caused by querying completion candidates for each character insert or deletion contrary to a timer-based approach of corfu. Querying of the backends is done in an interruptible manner (this part is based on corfu's internals) but heavy operations while searching for suggestion candidates can cause micro-stutters still.

Actually I think the paragraph is incorrect since `corfu-completion-overlay` and the usual Corfu popup with `corfu-auto=t` use the exactly same code to compute the candidates. If you observe micro stutter with `corfu-auto=t`, one should observe the same stutter with `corfu-completion-overlay`. 

In my experiment the stutter gets gradually better when increasing `corfu-auto-prefix` since then fewer candidates are generated by grep. The problem is not grep but Emacs itself which cannot handle many candidates well.

The only possible difference between `corfu-completion-overlay` and `corfu-auto` could be due to the child frame popup.

minad:
Hmm thinking about this more - is the difference that you don't compute the candidates again on each keypress but instead reuse the candidates computed in the beginning?

If this is the case, it would explain why your solution is faster. Unfortunately this is not correct since it breaks certain completion tables. But as far as I can tell it should work in the case of `cape-dict`.

Cape caching makes sure that `grep` is not called repeatedly, also with `corfu-auto=t`. There is still some recomputation taking place, namely the filtering of the candidates by the completion style via `all-completions` or `completion-all-completions`. This can be expensive if initially thousands of candidates were generated by `grep`. Therefore the problem is reduced for larger values of `corfu-auto-prefix`.

adamkruszewski:
> Actually I think the paragraph is incorrect since `corfu-completion-overlay` and the usual Corfu popup with `corfu-auto=t` use the exactly same code to compute the candidates. If you observe micro stutter with `corfu-auto=t`, one should observe the same stutter with `corfu-completion-overlay`.

I need to rephrase the paragraph -- "visually" it is seems *snappier* (less of stutter) **with** corfu-auto. That was the intention of the paragraph. Both use the same completion mechanism so it should be very similar but it maybe the popup redrawing makes it at least feel snappier?

> In my experiment the stutter gets gradually better when increasing `corfu-auto-prefix` since then fewer candidates are generated by grep. The problem is not grep but Emacs itself which cannot handle many candidates well.

Increasing corfu-auto-prefix helps -- I'll add that information. 

Grep or, when I think about it now, the size of the dict file was definitely the problem in my case. But I run Emacs inside a VM, which probably is to blame for that. When I have switched to completions based on your jinx package (thanks for another great package btw! reusing your Emacs module to libenchant - https://code.bsdgeek.org/adam/cape-jinx-completion) it definitely made my laptop fans not to spin uncontrollably like it was when using grep/dict file and the stuttering at least felt less of an issue. Number of completions look similar, if not larger even as I wasn't using a "full dictionary dump" with grep -- I only had roots of words dumped using aspell. With jinx under the hood I think I get more fine grained words completions now and is visibly faster. But as I have mentioned, running inside a VM does probably skew the results.

> The only possible difference between `corfu-completion-overlay` and `corfu-auto` could be due to the child frame popup.

Yes, it could be, the redrawing at least visually can make it *feel* different.

minad:
Okay, I understand. The updated paragraph makes more sense to me. 

As I wrote before in https://github.com/minad/corfu/pull/331#issuecomment-1554756040, the performance issue with an overlay directly behind the new input were my reason to not explore this further in #232. However the popup is visually more heavy and noisy, which can also add to the perception.

Regarding `cape-dict` - in my config I am experimenting with a combined `cape-dabbrev` and `cape-dict` backend. I don't use aggressive auto completion, so I don't notice significant performance issues. To mitigate possible performance issues one could pass a `--max-count` argument to grep. But `cape-dict` is disappointingly useless anyway as long as one doesn't use word lists which are sorted by frequency. Your Jinx/Enchant idea may help, however the purpose of Enchant is to provide corrected words and not expanded word for prefixes, so I worry that you will be missing a lot of useful expansions?

adamkruszewski:
Comparing to using a dumped dictionary it doesn't seem I lose too much of useful expansions. But the dictionary as you have pointed out is not sorted by the frequency of usage so it boils down to very similar set of expansions.
I mostly use it to just expand to the first one I see (i.e. I don't really use the popup list) and then write more to gradually expand or type to more or less what I want. It saves only few keystrokes, but give me some kind of immediate feedback I made a typo ;-) Not the ideal, but good enough. Making it suggest by frequency of use is a really good idea, but it would need to be done somewhere lower than elisp somehow as it would end up, after longer while, with quite a large dictionary itself - so back to how cape-dict works basically.

minad:
The idea of using your Jinx Capf an immediate typo correction sounds very good. The discussion here motivated me to also revise `cape-dict` a little bit. I use word frequency dictionaries from [here](https://wortschatz.uni-leipzig.de/de/download/English) in my setup, such that the more common words are listed earlier by grep. Then I added a new variable `cape-dict-limit` to limit the number of words, see https://github.com/minad/cape/commit/28a2a1655026ae68ed9180db6527694334c5d1d4. My hope is that this helps with both the performance and the usability issue. Independent of that I also investigate if it helps to limit the number of candidates on the level of the completion style filtering, see https://github.com/oantolin/orderless/issues/141. Obviously it is better if the candidate count is small in the first place, but such an additional limiter can help with completions like `M-x` or `C-h v`.

adamkruszewski:
Thanks! -- I'll play with ```cape-dict-limit``` over the weekend and will look at this sorted dictionary for English (need to look for a Polish sorted by frequency word list).

minad:
See https://wortschatz.uni-leipzig.de/de/download/Polish

PR #336, opened 2023-05-27
Add a section to use same key as insert separator and insertion

Cletip:


minad:
I meant that this code should be added to the [wiki](/minad/corfu/wiki), not the README. Can you please add it there?

PR #338, opened 2023-06-03
Create candidate preview before showing popup

axelf4:
The overlay previewing the current completion candidate may shift subsequent text, including overlays used for popups due to running in the terminal where child frames are unavailable. E.g.

     f|oobar long line that wr\
    aps +--------+
    |foo     |
    |>foobar |
    +--------+

where `|` is the cursor, and `\` indicates line wrapping and `>` the current candidate. Therefore it is necessary to create the candidate preview before positioning any completion popup overlays.

This commit also moves the call to `redisplay` into the child frame specific code, since it is redundant when overlays are used to display the popup, instead only causing flickering in the terminal.

This patch is small enough that I hope FSF copyright assignment should not be an issue.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -973,7 +973,8 @@ A scroll bar is displayed from LO to LO+BAR."
                              lines "\n"))
           (goto-char (point-min)))
         (setq corfu--frame (corfu--make-frame corfu--frame x y
-                                              width height (current-buffer)))))))
+                                              width height (current-buffer))))))
+  (redisplay 'force)) ;; XXX HACK Ensure that popup is redisplayed
 
 (cl-defgeneric corfu--popup-hide ()
   "Hide Corfu popup."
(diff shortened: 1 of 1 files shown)


minad:
Is this related to corfu-terminal? In principle I am okay with this - the order proposed in this PR is better, also because of the redisplay.

However unfortunately there is a reason why the code is as is. Displaying the overlay before the popup may break the popup position computation. I am not entirely sure which scenario was broken. Therefore I cannot merge this without much testing or further changes. Did you test this for a sufficiently long time (without corfu-terminal)?

axelf4:
> Is this related to corfu-terminal?

Yes, all overlay popups are broken with the current ordering, although I have only tested with [my own implementation](https://github.com/axelf4/dotfiles/blob/41f89d48f4661406d8c74e7f80fa1f05e7ba105c/.config/emacs/init.el#L571-L629).

> However unfortunately there is a reason why the code is as is.

I think the issue is as follows: When `corfu--preview-current` makes a zero-length overlay at `beg` with an `after-string` property, that causes the visual position of the `beg` buffer position to shift, even though the `after-string` should go *after* the position.  (This happens e.g. when completing with an empty prefix.) That seems like an Emacs bug to me?

Given this, I propose that `corfu--candidates-popup` be changed to instead take a `posn` computed before `corfu--preview-current`, i.e.

    (let (pos (posn-at-point (+ beg (length corfu--base))))
      (corfu--preview-current beg end)
      (corfu--candidates-popup pos))

(This is also simpler wrt. any dicey BIDI interactions I may have overlooked.) However it is a larger change than my original PR, and a semi-breaking change at that, so what do you think?

minad:
> Yes, all overlay popups are broken with the current ordering, although I have only tested with my own implementation.

Is corfu-terminal broken? I haven't heard of this before. cc @AkibAzmain

> my own implementation

If your own implementation doesn't make use of any third-party dependencies, we can add it here as a Corfu extension.

> That seems like an Emacs bug to me?

Yes, it is.

> Given this, I propose that corfu--candidates-popup be changed to instead take a posn computed before corfu--preview-current, i.e.

Sure, we can change the convention of `corfu--popup-show` to instead take a `posn` list as returned by `posn-at-point`.

> This patch is small enough that I hope FSF copyright assignment should not be an issue.

Given that this may require a larger change, are you willing to assign copyright for these contributions? Is it possible for you?

PR #345, opened 2023-07-05
add corfu-terminal

scturtle:
An over-simplified version of corfu-terminal + popon. Is such a thing mergeable?


minad:
No, since @AkibAzmain holds copyright on this code, while Corfu requires copyright assignment to the FSF. See https://github.com/minad/corfu#contributions. In principle I would like to see a terminal display as Corfu extension here, but I am not sure about the "over-simplified". From my personal experience, overlay-based implementations suffer from a great number of display issues and from time to time new issues appear even for Company. As an alternative to Popon you could try to port the overlay implementation from Company here.

PR #379, opened 2023-10-10
corfu--on: when (not foo) in global-corfu-modes, enable when not foo

mekeor:
I set `global-corfu-modes` to `'((not erc-modes))` and it stopped working anywhere. I'd expect this PR to fix this.

But I'm not sure if this is the right solution in general.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -1303,7 +1303,7 @@ symbols or elements of the form (not modes)."
                                   ('t t)
                                   ('nil 0)
                                   ((pred symbolp) (and (derived-mode-p p) t))
-                                  (`(not . ,m) (and (apply #'derived-mode-p m) 0)))))))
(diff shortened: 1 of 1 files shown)


minad:
I assume a final t is needed:

~~~ elisp
(setq global-corfu-modes '((not org-mode) t))
~~~

mekeor:
2023-10-10 22:13 ***@***.***:

> I assume a final t is needed:
>
> (setq global-corfu-modes '((not org-mode) t))

Oh right. Thank you very much.

I didn't know that `cl-loop`'s `thereis` clause will also stop when it's 0, i.e. non-nil.

> ‘thereis CONDITION’
>      This clause stops the loop when the specified form is 
>      non-‘nil’; in
>      this case, it returns that non-‘nil’ value.  If all the 
>      values were
>      ‘nil’, the loop returns ‘nil’.

PR #393, opened 2023-11-29
Add corfu-popup-location defcustom (above or below current line)

jrblevin:
When a completion backend proposes multi-line completions (e.g., codeium.el), the corfu popup (which appears below the point by default) covers the completion preview. This patch introduces a defcustom `corfu-popup-location` which allows the user to set a preference for the popup to appear above the current line if there is space available. This allows one to see both the completions and the preview in the case of a multi-line completion.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -67,6 +67,13 @@ The value should lie between 0 and corfu-count/2."
   "Popup maximum width in characters."
   :type 'natnum)
 
+(defcustom corfu-popup-location 'below
+  "Location of the Corfu popup.
+The value can be either 'above or 'below.  The default value is 'below."
+  :type '(choice (const :tag "Above the point" above)
+                 (const :tag "Below the point" below))
+  :group 'corfu)
+
 (defcustom corfu-cycle nil
   "Enable cycling for `corfu-next' and `corfu-previous'."
   :type 'boolean)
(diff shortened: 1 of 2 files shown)


minad:
Thanks. I would be happy to add an option for a preferred popup location. Did you sign the FSF papers, see https://github.com/minad/corfu#contributions?

corfu.el
```diff
@@ -972,9 +979,11 @@ A scroll bar is displayed from LO to LO+BAR."
              (x (max 0 (min (+ (car edge) (- (or (car pos) 0) ml (* cw off) border))
                             (- (frame-pixel-width) width))))
              (yb (+ (cadr edge) (window-tab-line-height) (or (cdr pos) 0) lh))
-             (y (if (> (+ yb (* corfu-count ch) lh lh) (frame-pixel-height))
+             (y (if (eq corfu-popup-location 'above)
                     (- yb height lh border border)
```
minad, line 985: I think we also have to check here if there is enough space above the input line. Currently the code first checks the space above, and if not sufficient moves the popup above. The option would ideally invert this behavior, first check above, and so on.

jrblevin:
> Thanks. I would be happy to add an option for a preferred popup location. Did you sign the FSF papers, see https://github.com/minad/corfu#contributions?

I spent a while searching for how to file the papers, but came up short. Do you possibly have a link with instructions?

corfu.el
```diff
@@ -67,6 +67,15 @@ The value should lie between 0 and corfu-count/2."
   "Popup maximum width in characters."
   :type 'natnum)
 
+(defcustom corfu-popup-location 'below
```
minad, line 70: This could be a simple boolean `corfu-popup-below` with the default set to t. Or we could support more values `above`, `below`, `always-above`, `always-below`.

corfu.el
jrblevin: You're right, I'll work on that.

corfu.el
jrblevin: A boolean was my first thought, but I thought this might be more general in case someone comes up with some other interesting cases in the future. Let me know your preference.

minad:
> I spent a while searching for how to file the papers, but came up short. Do you possibly have a link with instructions?

Take a look at the CONTRIBUTE file in the Emacs source. Please let me know how it goes.

corfu.el
minad: If we implement two behaviors (prefer above or prefer below), I suggest to use a boolean. We could still extend the variable later if necessary.

minad:
@jrblevin Ping. Any news here?

minad:
@jrblevin Please let me know if you plan to pick this up again.

PR #401, opened 2023-12-22
Support displaying the popup when there is only one exact match

galeo:
As discussed in #396, a `show` value is added to `corfu-on-exact-match`, and it handles both auto and manual completion.

Please have a look. Thanks.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -73,7 +73,7 @@ The value should lie between 0 and corfu-count/2."
 
 (defcustom corfu-on-exact-match 'insert
   "Configure how a single exact match should be handled."
-  :type '(choice (const insert) (const quit) (const nil)))
+  :type '(choice (const insert) (const show) (const quit) (const nil)))
 
 (defcustom corfu-continue-commands
   ;; nil is undefined command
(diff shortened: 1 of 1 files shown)


corfu.el
```diff
@@ -907,10 +908,12 @@ Lookup STR in CANDS to restore text properties."
...
              ;; If completion is finished and cannot be further completed,
+             ;; and the value of corfu-on-exact-match is not 'show,
              ;; return 'finished. Otherwise setup the Corfu popup.
-             (if (consp (completion-try-completion
-                         newstr table pred newpt
-                         (completion-metadata newstr table pred)))
+             (if (or (consp (completion-try-completion
+                             newstr table pred newpt
+                             (completion-metadata newstr table pred)))
+                     (eq corfu-on-exact-match 'show))
```
minad, line 916: Check corfu-on-exact-match before completion-try-completion for efficiency.

minad:
Thanks, this looks simple enough! I will give it a try.

corfu.el
```diff
@@ -889,12 +889,13 @@ Lookup STR in CANDS to restore text properties."
...
-           str 'finished
-           (alist-get 'corfu--candidates (corfu--recompute str pt table pred)))
-          t)
+      ((and (guard (not (eq corfu-on-exact-match 'show))) 't)
+       (goto-char end)
+       (corfu--message "Sole match")
+       (corfu--exit-function
+        str 'finished
+        (alist-get 'corfu--candidates (corfu--recompute str pt table pred)))
+       t)
```
minad, line 898: The behavior here does not seem to be right. When I complete manually and get an exact match (with corfu-on-exact-match=show), this branch is not taken, but also not the next one. The popup is not shown as it should.

galeo: Do you set `completion-cycle-threshold` to `nil`?

minad: Shouldn't matter?

galeo: Sorry. I misunderstood what you meant and the code has been updated to fix this issue.

minad: Thanks, looks all good now.

minad:
Merged. I simplified the setup slightly in https://github.com/minad/corfu/commit/9cebcfedca0ab165fcfc8a56acc2d2513a02b9e3. Please give it a check if everything works as expected for you.

galeo:
Did some quick testing and it works well, thanks!

PR #455, opened 2024-05-02
Close completion popup if text len is less than corfu-auto-prefix

krydos:
Currently the auto completion popup appears as soon as user enters `corfu-auto-prefix` number of characters but it never goes away even if user removes some characters. It may not sound like a problem but it freezes Emacs if there are too many options for completion. 

For example (when `corfu-auto-prefix` == `3`):
- Enter `(mes` (I want to enter `(message "...")` but then I changed my mind and want to enter something else)
- Completion popup appears
- Delete `s`
- Completion popup is still there even though the length of entered text is less than `corfu-auto-prefix`
- Delete `e` 
- Completion popup is still there and (at least in my case) Emacs is getting frozen for a few seconds because there are too many options available for letter `m` to auto complete. 

Expected behavior for me is to close the completion popup as soon as the number of entered characters are less than `corfu-auto-prefix`.

This issue is less prominent if you have a habit to use`M-Backspace` instead of `Backspace-Many-Times` as I used to do. 
I suffer from this issue every day but I can't find anyone complaining about it which makes me think that I either missed a variable that actually fixes this issue or I maybe have too many completion sources which makes my Emacs slow... 

In any case, let me know if you find this change valuable and let me know if I can improve anything code-wise.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -979,8 +979,9 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
       (`(,fun ,beg ,end ,table . ,plist)
        (let ((completion-in-region-mode-predicate
               (lambda ()
-                (when-let ((newbeg (car-safe (funcall fun))))
-                  (= newbeg beg))))
+                (and (<= corfu-auto-prefix (- (point) beg))
+                    (when-let ((newbeg (car-safe (funcall fun))))
+                      (= newbeg beg)))))
              (completion-extra-properties plist))
          (corfu--setup beg end table (plist-get plist :predicate))
          (corfu--exhibit 'auto))))))


minad:
Thanks for this proposal. It seems reasonable, but I am a little worried to change this behavior now. Also if you trigger manual completion, the popup won't go away again when deleting input. The popup must be dismissed explicitly with C-g or ESC.

In your comment there are a few red flags, regarding this feature request, which sound like a problematic configuration on your side:

> ...it freezes Emacs if there are too many options for completion...

> ...or I maybe have too many completion sources which makes my Emacs slow...

> ...I can't find anyone complaining about it...

Corfu should not freeze your Emacs. Corfu interrupts completion sources if they take too long and it is of course not recommended to use too many slow sources or sources creating tons of useless candidates. Can you please try to fix this, by reducing the amount of sources?

There are two other possibilities to resolve this:

1. An alternative completion style which does not return candidates for short input. However this interferes with manual completion.

~~~ elisp
(defun short-all-completions (string table pred point)
  (and (length< string 4)
       (completion-emacs21-all-completions string table pred point)))

(defun short-try-completion (string table pred point)
  (and (length< string 4)
       (completion-emacs21-try-completion string table pred point)))

(add-to-list 'completion-styles-alist
             '(short
               short-try-completion
               short-all-completions
               "Basic completion with length limit for speed up."))
~~~

2. Rebind DEL to another command which dismisses the popup for short input.

krydos:
Thanks for such a quick reply @minad 
I just went through my config and noticed that I have some `cape-*` functions added to `completion-at-point-functions` which may cause the slowness and apparently if no one else have this issue then it's not a package problem.

I can't also reproduce it with `emacs -Q`

I'll fix it on my end. 

Thank you for suggestions on how to fix it!

PR #508, opened 2024-10-19
Calculate popup width for lines with uneven character widths correctly

wyuenho:
# Introduction

Historically there seems to be a number of alignment problems with Corfu due to differences of font widths in different substrings within the completion strings, this often clash with the desire to align the prompt right below the beginning of the string to complete, which resulted in some unusual function signatures and yet the problem still remains.

## Problem 1
Screenshot 2024-10-19 at 5 38 06 PM

When using `corfu-margin-formatters` or `:affixation-function`, the returned icon may be an image that has a different width from the buffer default font width. Currently corfu simply multiplies the maximum string widths calculated as the number of columns required in `corfu--format-candidates`, and the `default-font-width` of the buffer together. This assumes every glyph in the strings has the same width, which may not be the case when customizing any of `corfu-margin-formatters`, `:affixation-function` or `:annotation-function`. The result is the final width calculation, which is in pixels, is off by the difference of the number of pixels between the icon image widths and the default font widths multiplied by the number of glyphs in these different widths, resulting in the annotation being cut off.

## Problem 2
Screenshot 2024-10-19 at 5 40 33 PM

In order to align the prompt to the beginning of the string to complete,  a prefix width denoted in the number of columns is used to calculate an offset, similar to problem 1, results in the wrong popup width calculation, so the prefix offset is also off by the difference between the default font width and the pixel width of the actual images.

## Problem 3
Screenshot 2024-10-19 at 5 43 26 PM

`corfu-left-margin-width` is set to 0 regardless of whether the curried margin formatter function actually returned a non-empty string, which also leads to the left margin of the popup being removed.

## Problem 4
Screenshot 2024-10-19 at 5 46 52 PM

If the completion candidate itself contains an emoji, similar to problem 1, in addition to the width of the candidate string is wrong, now even the height is wrong (nothing we can do about the pixel height tho, it's an Emacs limitation for emacs-devel to solve).

## Problem 5

Overly long strings are truncated without ellipses to indicate truncation, and/or the padding between the candidate and the suffix are uneven.

Screenshot 2024-10-19 at 5 49 18 PM
Screenshot 2024-10-19 at 5 50 19 PM
Screenshot 2024-10-19 at 5 50 32 PM

Even after truncating the suffix, the completion candidate string itself + prefix can still be longer than max-width, but the candidate string is not truncated.

Screenshot 2024-10-19 at 5 50 02 PM

## Problem 6

Padding is wrong if any of prefix, candidate or suffix use a variable font.

Screenshot 2024-10-19 at 5 56 58 PM

## Solution

Let `corfu--affixate` apply the margin formatters / affixation function / annotation function only. Return only a list of triples of strings.

Let `corfu--format-candidates` format, pad, and truncate the completion strings by pixels. Candidate strings and suffixes are truncated individually with pixel-based padding in between. Return only the final list of formatted strings.

Recalculate the string pixel widths for the prefix and the completion strings before calling `corfu--popup-show`, which should calculate the final pixel widths needed for the popup using the passed in prefix and candidate widths, now in pixels, directly without pretending every character in the strings having uniform widths.

Recalculate truncations and padding for the current candidate as a customized `corfu-current` face may also affect pixel widths.

## Final result

Screenshot 2024-10-19 at 6 00 49 PM
Screenshot 2024-10-19 at 6 01 14 PM
Screenshot 2024-10-19 at 6 01 46 PM
Screenshot 2024-10-19 at 6 01 56 PM
Screenshot 2024-10-19 at 6 03 04 PM
Screenshot 2024-10-19 at 6 04 48 PM
Screenshot 2024-10-19 at 6 05 43 PM
Screenshot 2024-10-19 at 6 07 09 PM
Screenshot 2024-10-19 at 6 07 27 PM
Screenshot 2024-10-19 at 6 07 45 PM

## Follow up

corfu-terminal will be broken by this PR, but I'll file a PR over there.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -775,16 +775,18 @@ buffers."
                     (setf s (replace-regexp-in-string "[ \t]*\n[ \t]*" " " s))))
   cands)
 
-(defun corfu--format-candidates (cands curr)
+(defun corfu--format-candidates (cands curr ml mr)
   "Format annotated CANDS.
-CURR is index of the currently selected candidate."
+CURR is index of the currently selected candidate.
+ML is the left margin padding in pixels on graphical displays or columns on the
+terminal.
+MR is the left margin padding in pixels on graphical displays or columns on the
+terminal."
   (let* ((cw (string-pixel-width (string-join (cl-loop for x in cands collect (car x)) "\n")))
          (pw (string-pixel-width (string-join (cl-loop for x in cands collect (cadr x)) "\n")))
          (sw (string-pixel-width (string-join (cl-loop for x in cands collect (caddr x)) "\n")))
          (fw (default-font-width))
          (width (max (+ pw cw sw) (* fw corfu-min-width)))
-         (ml (max 0 (ceiling (* fw corfu-left-margin-width))))
-         (mr (max 0 (ceiling (* fw corfu-right-margin-width))))
          (marginl (and (> ml 0) (propertize " " 'display `(space :width (,ml)))))
          (marginr (and (> mr 0) (propertize " " 'display `(space :width (,mr))))))
 
@@ -834,11 +836,16 @@ CURR is index of the currently selected candidate."
                (rcands (corfu--replace-newlines dcands))
                (prefix-pixel-width (string-pixel-width
                                     (string-join (cl-loop for c in rcands collect (cadr c)) "\n")))
+               (fw (default-font-width))
                ;; Disable the left margin if there are prefixes
-               (corfu-left-margin-width (if (> prefix-pixel-width 0) 0 corfu-left-margin-width))
-               (left-margin-width (max 0 (ceiling (* (default-font-width) corfu-left-margin-width))))
-               (offset (+ prefix-pixel-width left-margin-width))
-               (lines (corfu--format-candidates rcands curr))
+               (ml (if (> prefix-pixel-width 0) 0 corfu-left-margin-width))
+               (ml (max 0 (ceiling (* fw ml))))
+               ;; Adjust right margin width according to scroll bar width
+               (bw (max 0 (ceiling (* fw corfu-bar-width))))
+               (mr (max 0 (ceiling (* fw corfu-right-margin-width))))
+               (mr (- (max mr bw) (min mr bw)))
+               (offset (+ prefix-pixel-width ml))
+               (lines (corfu--format-candidates rcands curr ml mr))
                (content-width (string-pixel-width (string-join lines "\n"))))
     ;; Nonlinearity at the end and the beginning
     (when (/= corfu--scroll 0)
@@ -1084,8 +1091,9 @@ A scroll bar is displayed from LO to LO+BAR."
     (with-current-buffer (corfu--make-buffer " *corfu*")
       (let* ((ch (default-line-height))
              (cw (default-font-width))
-             (mr (ceiling (* cw corfu-right-margin-width)))
-             (bw (ceiling (* cw corfu-bar-width)))
+             (mr (max 0 (ceiling (* cw corfu-right-margin-width))))
+             (bw (max 0 (ceiling (* cw corfu-bar-width))))
+             (bw (min mr bw))
              (sbar (propertize " " 'display
                                (if corfu-bar-on-fringe
                                    '(right-fringe corfu-scroll-bar corfu-bar)
@@ -1094,7 +1102,7 @@ A scroll bar is displayed from LO to LO+BAR."
              (pos (posn-x-y pos))
              (width (+
                      ;; scroll bar
-                     (if lo (if corfu-bar-on-fringe (- mr bw) cw) 0)
+                     (if lo (if corfu-bar-on-fringe bw cw) 0)
                      ;; -4 because of margins and some additional safety
                      (min (* cw (min (- (frame-width) 4) corfu-max-width)) content-width)))
              ;; XXX HACK: Minimum popup height must be at least 1 line of the
--- a/corfu.el
+++ b/corfu.el
@@ -776,7 +776,7 @@ terminal."
              do
              ;; `corfu-current' may affect frame-width too
              (when (= i curr)
-               (cl-loop for s in (list cand prefix suffix)
+               (cl-loop for s in (list cand prefix suffix marginl marginr)
                         do (add-face-text-property 0 (length s) 'corfu-current t s)))
 
              (cl-incf i)
--- a/corfu.el
+++ b/corfu.el
@@ -736,7 +736,7 @@ FRAME is the existing frame."
 WIDTH is in pixels. If the string is longer than width when
 rendered, it is truncated with the last character(s) replaced
 with the result of `truncate-string-ellipsis'. If shorter,
-returns an empty string."
+returns the string unmodified."
   (let* ((glyphs (string-glyph-split str))
          (glyph-width (string-pixel-width (car glyphs)))
          (face (and glyphs (get-text-property 0 'face (car (last glyphs)))))
@@ -761,10 +761,24 @@ returns an empty string."
 
 (defun corfu--format-candidates (cands curr)
   "Format annotated CANDS.
-CURR is the currently selected candidate."
-  (let* ((cw (cl-loop for x in cands maximize (string-pixel-width (car x))))
-         (pw (cl-loop for x in cands maximize (string-pixel-width (cadr x))))
-         (sw (cl-loop for x in cands maximize (string-pixel-width (caddr x))))
+CURR is index of the currently selected candidate."
+  (let* ((cw 0)
+         (pw 0)
+         (sw 0)
+         (cands (cl-loop for (c p s) in cands
+                         with triple = nil
+                         do
+                         (let (cpw ppw spw)
+                           (setq cpw (string-pixel-width c)
+                                 ppw (string-pixel-width p)
+                                 spw (string-pixel-width s)
+                                 cw (max cw cpw)
+                                 pw (max pw ppw)
+                                 sw (max sw spw)
+                                 triple (list (cons cpw c)
+                                              (cons ppw p)
+                                              (cons spw s))))
+                         collect triple))
          (width (+ pw cw sw))
          (fw (default-font-width))
          (min-width (* corfu-min-width fw))
@@ -782,7 +796,7 @@ CURR is the currently selected candidate."
     (when (< width min-width)
       (setq cw (+ cw (- min-width width))))
 
-    (cl-loop for (cand prefix suffix) in cands
+    (cl-loop for ((cand-width . cand) (prefix-width . prefix) (suffix-width . suffix)) in cands
              with i = 0
              do
              ;; `corfu-current' may affect width too
@@ -790,8 +804,13 @@ CURR is the currently selected candidate."
                (cl-loop for s in (list cand prefix suffix)
                         do (add-face-text-property 0 (length s) 'corfu-current t s)))
 
-             (setq cand (corfu--truncate-string-to-pixel-width cand cw)
-                   suffix (corfu--truncate-string-to-pixel-width suffix sw))
+             (when (> cand-width cw)
+               (setq cand (corfu--truncate-string-to-pixel-width cand cw)
+                     cand-width (string-pixel-width cand)))
+
+             (when (> suffix-width sw)
+               (setq suffix (corfu--truncate-string-to-pixel-width suffix sw)
+                     suffix-width (string-pixel-width suffix)))
 
              (cl-incf i)
 
@@ -799,10 +818,10 @@ CURR is the currently selected candidate."
              ;; do not use `:align-to' as it is buggy
              (concat
               prefix
-              (propertize " " 'display `(space :width (,(- pw (string-pixel-width prefix)))))
+              (propertize " " 'display `(space :width (,(- pw prefix-width))))
               cand
-              (propertize " " 'display `(space :width (,(- cw (string-pixel-width cand)))))
-              (propertize " " 'display `(space :width (,(- sw (string-pixel-width suffix)))))
+              (propertize " " 'display `(space :width (,(- cw cand-width))))
+              (propertize " " 'display `(space :width (,(- sw suffix-width))))
               suffix))))
 
 (defun corfu--compute-scroll ()
--- a/corfu.el
+++ b/corfu.el
@@ -136,6 +136,18 @@ separator: Only stay alive if there is no match and
   "Width of the right margin in units of the character width."
   :type 'float)
 
+(defcustom corfu-bar-width 0.2
+  "Width of the bar in units of the character width."
+  :type 'float)
+
+(defcustom corfu-bar-on-fringe t
+  "Whether the bar is shown on the fringe.
+
+If t, the bar is shown on the fringe with no truncation indicator.
+If nil, the bar is shown on the margin with ellipsis as the truncation
+indicator."
+  :type 'boolean)
+
 (defcustom corfu-margin-formatters nil
   "Registry for margin formatter functions.
 Each function of the list is called with the completion metadata as
@@ -207,8 +219,10 @@ See also the settings `corfu-auto-delay', `corfu-auto-prefix' and
   "Face used to highlight the currently selected candidate.")
 
 (defface corfu-bar
-  '((((class color) (min-colors 88) (background dark)) :background "#a8a8a8")
-    (((class color) (min-colors 88) (background light)) :background "#505050")
+  '((((class color) (min-colors 88) (background dark))
+     :background "#a8a8a8" :foreground "#a8a8a8")
+    (((class color) (min-colors 88) (background light))
+     :background "#505050" :foreground "#505050")
     (t :background "gray"))
   "The background color is used for the scrollbar indicator.")
 
@@ -225,6 +239,27 @@ See also the settings `corfu-auto-delay', `corfu-auto-prefix' and
 (defface corfu-deprecated
   '((t :inherit shadow :strike-through t))
   "Face used for deprecated candidates.")
+
+(define-fringe-bitmap 'corfu-scroll-bar
+  (vector
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   )
+  255 16 '(center . t))
 
 (defvar-keymap corfu-mode-map
   :doc "Keymap used when `corfu-mode' is active.")
@@ -411,6 +446,10 @@ length override, set to t for manual completion."
       (setq-local face-remapping-alist (copy-tree fr)
                   line-spacing ls)
       (cl-pushnew 'corfu-default (alist-get 'default face-remapping-alist))
+      ;; truncation ellipsis when `corfu-bar-on-fringe' is nil
+      (let ((dt (make-display-table)))
+        (set-display-table-slot dt 'truncation ?\x2026)
+        (setq-local buffer-display-table dt))
       buffer)))
 
 (defvar x-gtk-resize-child-frames) ;; Not present on non-gtk builds
@@ -1045,12 +1084,19 @@ A scroll bar is displayed from LO to LO+BAR."
     (with-current-buffer (corfu--make-buffer " *corfu*")
       (let* ((ch (default-line-height))
              (cw (default-font-width))
-             (sbar (propertize " " 'display `((margin right-margin)
-                                              ,(propertize " " 'face 'corfu-bar))))
+             (mr (ceiling (* cw corfu-right-margin-width)))
+             (bw (ceiling (* cw corfu-bar-width)))
+             (sbar (propertize " " 'display
+                               (if corfu-bar-on-fringe
+                                   '(right-fringe corfu-scroll-bar corfu-bar)
+                                 `((margin right-margin)
+                                   ,(propertize " " 'face 'corfu-bar)))))
              (pos (posn-x-y pos))
-             (width (+ (if lo cw 0) ;; scroll bar margin
-                       ;; -4 because of margins and some additional safety
-                       (min (* cw (min (- (frame-width) 4) corfu-max-width)) content-width)))
+             (width (+
+                     ;; scroll bar
+                     (if lo (if corfu-bar-on-fringe (- mr bw) cw) 0)
+                     ;; -4 because of margins and some additional safety
+                     (min (* cw (min (- (frame-width) 4) corfu-max-width)) content-width)))
              ;; XXX HACK: Minimum popup height must be at least 1 line of the
              ;; parent frame (gh:minad/corfu#261).
              (height (max lh (* (length lines) ch)))
@@ -1064,14 +1110,14 @@ A scroll bar is displayed from LO to LO+BAR."
                   yb)))
         (with-silent-modifications
           (erase-buffer)
+
           ;; scroll bar required
-          (setq-local right-margin-width (if lo 1 0))
-          ;; truncation required
-          (if (> content-width width)
-              (let ((dt (make-display-table)))
-                (set-display-table-slot dt 'truncation ?\x2026)
-                (setq-local buffer-display-table dt))
-            (setq-local buffer-display-table nil))
+          (if lo
+              (if corfu-bar-on-fringe
+                  (setq-local right-fringe-width bw)
+                (setq-local right-margin-width 1))
+            (setq-local right-fringe-width 0)
+            (setq-local right-margin-width 0))
 
           (insert (string-join
                    (cl-loop for i from 0 to (1- (length lines))
--- a/corfu.el
+++ b/corfu.el
@@ -219,10 +219,8 @@ See also the settings `corfu-auto-delay', `corfu-auto-prefix' and
   "Face used to highlight the currently selected candidate.")
 
 (defface corfu-bar
-  '((((class color) (min-colors 88) (background dark))
-     :background "#a8a8a8" :foreground "#a8a8a8")
-    (((class color) (min-colors 88) (background light))
-     :background "#505050" :foreground "#505050")
+  '((((class color) (min-colors 88) (background dark)) :background "#a8a8a8")
+    (((class color) (min-colors 88) (background light)) :background "#505050")
     (t :background "gray"))
   "The background color is used for the scrollbar indicator.")
 
@@ -241,25 +239,8 @@ See also the settings `corfu-auto-delay', `corfu-auto-prefix' and
   "Face used for deprecated candidates.")
 
 (define-fringe-bitmap 'corfu-scroll-bar
-  (vector
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   #b11111111111111
-   )
-  255 16 '(center . t))
+  (vector #b0)
+  255 16 '(center t))
 
 (defvar-keymap corfu-mode-map
   :doc "Keymap used when `corfu-mode' is active.")
--- a/corfu.el
+++ b/corfu.el
@@ -36,7 +36,6 @@
 ;;; Code:
 
 (require 'compat)
-(require 'mule-util)
 (eval-when-compile
   (require 'cl-lib)
   (require 'subr-x))
@@ -135,10 +134,6 @@ separator: Only stay alive if there is no match and
 
 (defcustom corfu-right-margin-width 0.5
   "Width of the right margin in units of the character width."
-  :type 'float)
-
-(defcustom corfu-bar-width 0.2
-  "Width of the bar in units of the character width."
   :type 'float)
 
 (defcustom corfu-margin-formatters nil
@@ -433,9 +428,11 @@ length override, set to t for manual completion."
          'resize-mode)))
 
 ;; Function adapted from posframe.el by tumashu
-(defun corfu--make-frame (frame x y width height buffer)
+(defun corfu--make-frame (frame x y width height buffer &optional show-special-glyphs)
   "Show BUFFER in child frame at X/Y with WIDTH/HEIGHT.
-FRAME is the existing frame."
+FRAME is the existing frame.
+If optional SHOW-SPECIAL-GLYPHS is non-nil, truncation indicators are allowed in
+buffers."
   (when-let (((frame-live-p frame))
              (timer (frame-parameter frame 'corfu--hide-timer)))
     (cancel-timer timer)
@@ -483,9 +480,10 @@ FRAME is the existing frame."
            (diff (cl-loop for p in should for (k . v) = p
                           unless (equal (alist-get k is) v) collect p)))
       (when diff (modify-frame-parameters frame diff)))
+    (set-frame-parameter frame 'no-special-glyphs (not show-special-glyphs))
     (let ((win (frame-root-window frame)))
-      (unless (eq (window-buffer win) buffer)
-        (set-window-buffer win buffer))
+      ;; always set buffer in order to update margin widths
+      (set-window-buffer win buffer)
       ;; Disallow selection of root window (gh:minad/corfu#63)
       (set-window-parameter win 'no-delete-other-windows t)
       (set-window-parameter win 'no-other-window t)
@@ -497,6 +495,13 @@ FRAME is the existing frame."
       (unless (and (= x px) (= y py))
         (set-frame-position frame x y))))
   (make-frame-visible frame)
+  ;; If the buffer had never been shown before, the margin text will not be
+  ;; visible until the frame is visible, so we need to force the window to
+  ;; update again
+  (with-current-buffer buffer
+    (when (= buffer-display-count 1)
+      (let ((win (frame-root-window frame)))
+        (set-window-buffer win buffer))))
   ;; Unparent child frame if EXWM is used, otherwise EXWM buffers are drawn on
   ;; top of the Corfu child frame.
   (when (and (bound-and-true-p exwm--connection) (frame-parent frame))
@@ -731,98 +736,42 @@ FRAME is the existing frame."
                     (setf s (replace-regexp-in-string "[ \t]*\n[ \t]*" " " s))))
   cands)
 
-(defun corfu--truncate-string-to-pixel-width (str width)
-  "Truncate string STR to WIDTH.
-WIDTH is in pixels. If the string is longer than width when
-rendered, it is truncated with the last character(s) replaced
-with the result of `truncate-string-ellipsis'. If shorter,
-returns the string unmodified."
-  (let* ((glyphs (string-glyph-split str))
-         (glyph-width (string-pixel-width (car glyphs)))
-         (face (and glyphs (get-text-property 0 'face (car (last glyphs)))))
-         (ellipsis (apply 'propertize (truncate-string-ellipsis) (if face `(face ,face))))
-         (ellipsis-width (string-pixel-width ellipsis))
-         result)
-    (while (and glyphs (<= glyph-width width))
-      (setq result (cons (car glyphs) result)
-            glyphs (cdr glyphs)
-            width (- width glyph-width)
-            glyph-width (string-pixel-width (car glyphs))))
-
-    (when (and glyphs result) ;; truncated
-      (while (and result (> ellipsis-width width))
-        (setq glyphs (cons (car result) glyphs)
-              result (cdr result)
-              width (+ width glyph-width)
-              glyph-width (string-pixel-width (car result))))
-      (setq result (cons ellipsis result)))
-
-    (string-join (nreverse result))))
-
 (defun corfu--format-candidates (cands curr)
   "Format annotated CANDS.
 CURR is index of the currently selected candidate."
-  (let* ((cw 0)
-         (pw 0)
-         (sw 0)
-         (cands (cl-loop for (c p s) in cands
-                         with triple = nil
-                         do
-                         (let (cpw ppw spw)
-                           (setq cpw (string-pixel-width c)
-                                 ppw (string-pixel-width p)
-                                 spw (string-pixel-width s)
-                                 cw (max cw cpw)
-                                 pw (max pw ppw)
-                                 sw (max sw spw)
-                                 triple (list (cons cpw c)
-                                              (cons ppw p)
-                                              (cons spw s))))
-                         collect triple))
-         (width (+ pw cw sw))
+  (let* ((cw (string-pixel-width (string-join (cl-loop for x in cands collect (car x)) "\n")))
+         (pw (string-pixel-width (string-join (cl-loop for x in cands collect (cadr x)) "\n")))
+         (sw (string-pixel-width (string-join (cl-loop for x in cands collect (caddr x)) "\n")))
          (fw (default-font-width))
-         (min-width (* corfu-min-width fw))
-         ;; -4 because of margins and some additional safety
-         (line-width (* (min corfu-max-width (- (frame-width) 4)) fw)))
-    ;; ensure line width is > min width and < max width, no need to ensure
-    ;; there's at least 1 space between segments, capf should have taken cared
-    ;; of it.
-    (when (> width line-width)
-      (setq sw (max 0 (- line-width pw cw))
-            width (+ pw cw sw))
-      (when (> width line-width)
-        (setq cw (max 0 (- line-width pw))
-              width (+ pw cw sw))))
-    (when (< width min-width)
-      (setq cw (+ cw (- min-width width))))
+         (width (max (+ pw cw sw) (* fw corfu-min-width)))
+         (ml (max 0 (ceiling (* fw corfu-left-margin-width))))
+         (mr (max 0 (ceiling (* fw corfu-right-margin-width))))
+         (marginl (and (> ml 0) (propertize " " 'display `(space :width (,ml)))))
+         (marginr (and (> mr 0) (propertize " " 'display `(space :width (,mr))))))
 
-    (cl-loop for ((cand-width . cand) (prefix-width . prefix) (suffix-width . suffix)) in cands
+    (cl-loop for (cand prefix suffix) in cands
              with i = 0
              do
-             ;; `corfu-current' may affect width too
+             ;; `corfu-current' may affect frame-width too
              (when (= i curr)
                (cl-loop for s in (list cand prefix suffix)
                         do (add-face-text-property 0 (length s) 'corfu-current t s)))
 
-             (when (> cand-width cw)
-               (setq cand (corfu--truncate-string-to-pixel-width cand cw)
-                     cand-width (string-pixel-width cand)))
-
-             (when (> suffix-width sw)
-               (setq suffix (corfu--truncate-string-to-pixel-width suffix sw)
-                     suffix-width (string-pixel-width suffix)))
-
              (cl-incf i)
 
+             ;; do not use relative space pixel params as they may lead to wrong
+             ;; `string-pixel-width' results
              collect
-             ;; do not use `:align-to' as it is buggy
              (concat
+              marginl
               prefix
-              (propertize " " 'display `(space :width (,(- pw prefix-width))))
+              (propertize " " 'display `(space :align-to (,(+ ml pw))))
               cand
-              (propertize " " 'display `(space :width (,(- cw cand-width))))
-              (propertize " " 'display `(space :width (,(- sw suffix-width))))
-              suffix))))
+              (propertize " " 'display `(space :align-to (,(+ ml pw cw
+                                                              (- width (+ pw cw sw))
+                                                              (- sw (string-pixel-width suffix))))))
+              suffix
+              marginr))))
 
 (defun corfu--compute-scroll ()
   "Compute new scroll position."
@@ -844,22 +793,21 @@ CURR is index of the currently selected candidate."
                                  for c in (nthcdr corfu--scroll corfu--candidates)
                                  collect (funcall corfu--hilit (substring c)))))
                (rcands (corfu--replace-newlines dcands))
-               (lines (corfu--format-candidates rcands curr))
-               ;; We can't naively multiply `default-font-width' with the max
-               ;; line widths in columns because the lines may contain
-               ;; propertized substrings that have different widths than the
-               ;; default
-               (prefix-pixel-width (cl-loop for c in rcands maximize (string-pixel-width (cadr c))))
-               (line-pixel-width (cl-loop for l in lines maximize (string-pixel-width l)))
+               (prefix-pixel-width (string-pixel-width
+                                    (string-join (cl-loop for c in rcands collect (cadr c)) "\n")))
                ;; Disable the left margin if there are prefixes
-               (corfu-left-margin-width (if (> prefix-pixel-width 0) 0 corfu-left-margin-width)))
+               (corfu-left-margin-width (if (> prefix-pixel-width 0) 0 corfu-left-margin-width))
+               (left-margin-width (max 0 (ceiling (* (default-font-width) corfu-left-margin-width))))
+               (offset (+ prefix-pixel-width left-margin-width))
+               (lines (corfu--format-candidates rcands curr))
+               (content-width (string-pixel-width (string-join lines "\n"))))
     ;; Nonlinearity at the end and the beginning
     (when (/= corfu--scroll 0)
       (setq lo (max 1 lo)))
     (when (/= last corfu--total)
       (setq lo (min (- corfu-count bar 2) lo)))
-    (corfu--popup-show pos prefix-pixel-width line-pixel-width lines
-                       curr (and (> corfu--total corfu-count) lo) bar)))
+    (corfu--popup-show pos offset content-width lines curr
+                       (and (> corfu--total corfu-count) lo) bar)))
 
 (defun corfu--range-valid-p ()
   "Check the completion range, return non-nil if valid."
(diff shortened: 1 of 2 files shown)


corfu.el
```diff
@@ -723,36 +724,94 @@ FRAME is the existing frame."
...
+  cands)
+
+(defun corfu--truncate-string-to-pixel-width (str width)
+  "Truncate string STR to WIDTH.
+WIDTH is in pixels. If the string is longer than width when
+rendered, it is truncated with the last character(s) replaced
+with the result of `truncate-string-ellipsis'. If shorter,
+returns an empty string."
+  (let* ((glyphs (string-glyph-split str))
+         (glyph-width (string-pixel-width (car glyphs)))
```
minad, line 741: Thanks for working on this. 

There is a small complication regarding `string-pixel-width`. It is only available on Emacs 29 and newer. The implementation of `string-pixel-width` is only efficient on Emacs 30 and newer. Therefore we must provide a back ported version `corfu--string-pixel-width`.

Furthermore I hope that this change does not break the child frame on terminal efforts by Gerd Möllman, which explicitly aims to support Corfu. This needs testing or we may have to hold merging this back for a while.

corfu.el
```diff
@@ -1151,7 +1224,8 @@ AUTO is non-nil when initializing auto completion."
         ('t nil)
         ('nil corfu--input)
         ('separator (seq-contains-p (car corfu--input) corfu-separator)))
-      (corfu--popup-show (posn-at-point beg) 0 8 '(#("No match" 0 8 (face italic)))))
+      (corfu--popup-show (posn-at-point beg) 0 (string-pixel-width no-match-msg)
```
minad, line 1227: Shouldn't the italic face be taken into account here?

corfu.el
```diff
@@ -1105,7 +1175,9 @@ A scroll bar is displayed from LO to LO+BAR."
              (when (and dep (funcall dep c))
                (setcar x (setq c (substring c)))
                (add-face-text-property 0 (length c) 'corfu-deprecated 'append c)))
-    (cons mf cands)))
+    cands))
+
+(define-obsolete-function-alias 'corfu--affixate 'corfu--decorate "1.6")
```
minad, line 1180: I assume you deprecated the function due to a breakage in behavior. But given that it is an internal function I would prefer to keep the name as is.

corfu.el
minad: Also there is the question of efficiency. The last time I tried this, string-pixel-width was quite slow in the order of milliseconds due to inefficiencies of the emacs display mechanism. This won't fly if we call it 20 or more times. Maybe the problem is not as severe as I remember or maybe we could compute the width of multiple strings in bulk if all we need is the maximum width. Did you run a benchmark on string-pixel-width or the whole result?

wyuenho: > The implementation of string-pixel-width is only efficient on Emacs 30 and newer.

[Both](https://github.com/emacs-mirror/emacs/blob/emacs-29/lisp/emacs-lisp/subr-x.el) [implementations](https://github.com/emacs-mirror/emacs/blob/emacs-30/lisp/emacs-lisp/subr-x.el) are exactly the same, namely both calls a C function `buffer-text-pixel-size`, which is also almost identical in [both](https://github.com/emacs-mirror/emacs/blob/emacs-29/src/xdisp.c) [versions](https://github.com/emacs-mirror/emacs/blob/emacs-30/src/xdisp.c). To emulate that in Elisp, one would have to spawn a hidden frame and insert the characters in the root window in the new frame to measure the width, this is super slow and sounds like a lot of work for compat instead of here.

> Furthermore I hope that this change does not break the child frame on terminal efforts by Gerd Möllman, which explicitly aims to support Corfu. This needs testing or we may have to hold merging this back for a while.

I'm not familiar with this effort or what it has to do with `string-pixel-width`. `string-pixel-width` just returns 1 per glyph on terminals, so everything reverts back to uniform character widths and the math still works out.

> The last time I tried this, string-pixel-width was quite slow in the order of milliseconds due to inefficiencies of the emacs display mechanism.

`string-pixel-width` is slightly slower than `string-width`, but not by much on my M1 MBP on the NS port, we are talking about a few milliseconds per 100 repetitions for a string with 100 glyphs. That's 10 million glyph measurements per second.

Screenshot 2024-10-20 at 5 02 18 AM

minad: > [Both](https://github.com/emacs-mirror/emacs/blob/emacs-29/lisp/emacs-lisp/subr-x.el) [implementations](https://github.com/emacs-mirror/emacs/blob/emacs-30/lisp/emacs-lisp/subr-x.el) are exactly the same, namely both calls a C function buffer-text-pixel-size, which is also almost identical in [both](https://github.com/emacs-mirror/emacs/blob/emacs-29/src/xdisp.c) [versions](https://github.com/emacs-mirror/emacs/blob/emacs-30/src/xdisp.c).

Oh, I misremembered this then. I saw that `string-pixel-width` unnecessarily copied the string until Emacs 31 and had minor issues with text properties.

> Let's just bump the minimum Emacs version required to 29 when Emacs 30 is released. I've never heard of people using an Emacs 2 major versions old.

I don't consider this an option just yet. I just bumped the required version to 28. I will bump to 29 when 31 is released.

> I'm not familiar with this effort or what it has to do with string-pixel-width. string-pixel-width just returns 1 per glyph on terminals, so everything reverts back to uniform character widths and the math still works out.

Okay that's good. Worth a test still.

> string-pixel-width is about the same as string-width on my M1 MBP on the NS port, we are talking about a few milliseconds per 100 repetitions for a string with 100 glyphs. That's 10 million glyph measurements per second.

I just tried your benchmark. On Linux `string-width` is at least 10 times slower. It should still be fast enough.

corfu.el
```diff
@@ -723,36 +724,94 @@ FRAME is the existing frame."
...
+             (concat
+              prefix
+              (apply 'propertize " " 'display `(space :width (,(- pw (string-pixel-width prefix))))
+                     (if prefix-face `(face ,prefix-face)))
+              cand
+              (apply 'propertize " " 'display `(space :width (,(- cw (string-pixel-width cand))))
+                     (if cand-face `(face ,cand-face)))
+              (apply 'propertize " " 'display `(space :width (,(- sw (string-pixel-width suffix))))
+                     (if suffix-face `(face ,suffix-face)))
+              suffix))))
```
minad, line 814: Why do we have to copy the face here? It doesn't seem necessary given that it will only affect the background of the spaces and the candidates and affixes won't have any. I generally prefer to go with the simplest solution that just works, so if we can remove this please do.

corfu.el
minad: Overall your patch looks good. I only have concerns about too many calls to `string-pixel-width`. In particular `corfu--truncate-string-to-pixel-width` doesn't look efficient enough because of the calls to `string-pixel-width` on every glyph if I understand correctly.

~~~ elisp
(benchmark-run 10000 (string-pixel-width "x")) -> 0.2s !!!
~~~

10000 is a bit too high, but we are entering performance relevant territory nevertheless. This means we have to save `string-pixel-width` calls.

corfu.el
wyuenho: Removing the faces just renders a single space in my testing. I've also tried `:align-to` to save 2 calls to `string-pixel-width` too, but somehow the result is off by a couple of pixels for the longest string when there's no prefix. Basically, this block is the only reliable way to pixel perfection I've found. What I've saying is, DO NOT change this unless exhaustive testing was done.

wyuenho: Nope, I just forgot the last argument to `apply` must be a list.

corfu.el
wyuenho: > 10000 is a bit too high, but we are entering performance relevant territory nevertheless. This means we have to save string-pixel-width calls.

Nah, 1000 calls is more realistic. 1000 calls is still ~0.01 seconds, that's enough for >60Hz FPS.

The only concern seems to be that `string-pixel-width` doesn't exist in Emacs < 29, but even Debian 12 already has Emacs 29.4 in the backport suite. When Debian 13 is released, we can safely bump the latest Emacs version requirement to 29.

minad: > Nah, 1000 calls is more realistic. 1000 calls is still ~0.01 seconds, that's enough for >60Hz FPS.

On my system 1000 calls is 0.02s which already corresponds to only 50 frames per second. And all this ONLY for the pixel alignment computation. The display part should be the cheaper part of the whole completion process, and not the one limiting the frame rate!

All I wanted to say is that we are in performance relevant territory here. Maybe the truncate string function could use some bisection technique (see also `eww--limit-string-pixelwise`) or it could approximate via `string-width`/`truncate-string-to-width`. I think approximating the truncation would be my preference, since it would be simpler and probably good enough. We would still get good alignment of the prefixes/candidate/suffixes via `string-pixel-width`.

wyuenho: What do you get from this?

```(benchmark-run 60 (corfu--format-candidates (cl-loop repeat 50 collect (copy-sequence '("" "prefix" "suffix"))) 0))```

Comment (minad) [reply corfu.el:741]: I haven't tried your version, but if I run the following code with `corfu--format-candidates` from the main branch I get 0.06s.

~~~ elisp
(benchmark-run 60 (corfu--format-candidates (cl-loop repeat 50 collect (list "candidatecandidatecandidatecandidate" "prefix" "suffixsuffixsuffixsuffixsuffixsuffix"))))
~~~

This means the whole formatting takes 1ms right now for a single frame which is quite fast. If we add the pixel formatting with the assumed 1000 calls we will add 20ms such that the formatting becomes 20 times more expensive than it is now.

> I can experiment with binary search a bit

I mentioned this above, if the approximate truncation works well enough, I would prefer that. It will be faster, simpler and probably good enough. Truncation doesn't have to be pixel perfect since we cannot make the cut at an exact width anyway.

Comment (minad) [reply corfu.el:741]: I haven't tried your version, but if I run the following code with `corfu--format-candidates` from the main branch I get 0.06s.

~~~ elisp
(benchmark-run 60 (corfu--format-candidates (cl-loop repeat 50 collect (list "candidatecandidatecandidatecandidate" "prefix" "suffixsuffixsuffixsuffixsuffixsuffix"))))
~~~

This means the whole formatting takes 1ms right now for a single frame which is quite fast. If we add the pixel formatting with the assumed 1000 calls we will add 20ms such that the formatting becomes 20 times more expensive than it is now.

> I can experiment with binary search a bit

I mentioned this above, if the approximate truncation works well enough, I would prefer that. It will be faster, simpler and probably good enough. Truncation doesn't have to be pixel perfect since we cannot make the cut at an exact width anyway.

Comment (wyuenho) [reply corfu.el:741]: d4c52235717991bfbad22b1b0c9016fa320104fc cuts down CPU times by ~20x for the usual case where no truncation is needed.  When there is, it's still faster by eliminating a couple of calls to `string-pixel-width`. This implementation takes the time back to around 0.12s for the above benchmark. Twice as slow as before but pixel perfect.

Comment (wyuenho) [reply corfu.el:741]: WRT to `string-pixel-width` not available on Emacs < 29, how about extracting this PR to an extension?

Comment (minad) [reply corfu.el:741]: > https://github.com/minad/corfu/commit/675c089a732927354ea6619a07d1eaf480188d86 cuts down CPU time by ~20x for the usual case where no truncation is needed.

Sounds good.

> WRT to string-pixel-width not available on Emacs < 29, how about extracting this PR to an extension? I can submit a separate PR to clean up HEAD to make the writing the extension a little easier.

I think this is not extension-style code. I am afraid we need some backward compatibility code. Also string-glyph-split is not available before 29.1.

It shouldn't be a problem to port back string-pixel-width - we can use the more efficient version from Emacs 31. For string-glyph-split a different solution is needed. I still wonder if an alternative truncation method would work (like eww--limit-string-pixelwise).

Comment (wyuenho) [reply corfu.el:741]: > It shouldn't be a problem to port back string-pixel-width - we can use the more efficient version from Emacs 31. For string-glyph-split a different solution is needed.

It should be the other way around. `string-glyph-split` can be backported easily all the way to Emacs 20, but let's not kid ourselves. Emacs didn't even support Emoji properly until Emacs 28.

`string-pixel-split` is the hard one because it needs to blitz the characters to an off screen window, of which you can only hack your way to it.

> I think this is not extension-style code. I am afraid we need some backward compatibility code.

I think I can clean up HEAD, and move this PR to an Emacs 29 + extension that uses some advice + cl-flet magic to get to pixel perfect.

Comment (minad) [reply corfu.el:741]: > I think I can clean up HEAD, and move this PR to an Emacs 29+ extension that uses some advice + cl-flet and/or cl-defgeneric magic to get to pixel perfect.

No, thank you. See above. I don't want to maintain arbitrary code in extensions. The extensions currently have a well defined purpose and don't just override existing code.

> string-pixel-width is the hard one because it needs to blitz the characters to an off screen window, of which you can only hack your way to it.

string-pixel-width can be ported back to 28. window-text-pixel-size has been available for a long time. The computation also works if the window is set to a temporary buffer without displaying it.

> It should be the other way around. string-glyph-split can be backported easily all the way to Emacs 20, but let's not kid ourselves. Emacs didn't even support Emoji properly until Emacs 28.

Yes, but my point is that porting back string-glyph-split is not necessary. We can use simpler logic for the truncation, while we definitely need the backport of string-pixel-width.

Comment (wyuenho) [reply corfu.el:741]: There seems to be some confusion about `string-glyph-split` and `string-pixel-width`, why they are necessary for truncation and whether they can be ported.

> Porting back string-glyph-split is not a necessity.

`string-glyph-split` is necessary for truncation because of [composed characters](https://elpa.gnu.org/packages/doc/compat.html#index-string_002dglyph_002dsplit) such as 👨‍⚖️.
    `(truncate-string-to-width "👨‍⚖️1" 3 0)` returns Screenshot 2024-10-21 at 1 59 16 PM in Emacs 29, and `(truncate-string-to-width "👨‍⚖️1" 2 0)` returns just `"👨"`, both results are wrong. Porting this function amounts to simply directly copying the [definition](https://github.com/emacs-mirror/emacs/blob/emacs-29/lisp/emacs-lisp/subr-x.el) to compat. (`find-composition-internal` has been available for decades). This should not be a hurdle. I don't understand why it was never ported to compat.
    
> string-pixel-width can be ported back to 28. window-text-pixel-size has been available for a long time. The computation also works if the window is set to a temporary buffer without displaying it.

I've chased back your [conversations](https://github.com/emacs-compat/compat/commit/7c17c96c1e0e98033481394e05b20fc3399a373c) around `string-pixel-width` in compat, it seems the issues were that an [earlier implementation](https://github.com/emacs-compat/compat/commit/7c17c96c1e0e98033481394e05b20fc3399a373c) that didn't turn off `display-line-number-mode` and unset `line-prefix` and `wrap-prefix` would sometimes return 0, and because of compat's version of `buffer-text-pixel-size` doesn't contain and can't implement [this bug fix](https://github.com/emacs-mirror/emacs/blob/emacs-30/src/xdisp.c#L11905-L11908), which incidentally also appears in Emacs 29's native version. The former issue is a mystery, and the later, just makes it functionally equivalent to Emacs 29 and irrelevant for Corfu. Clarification is needed on the former issue.

I've experimented with a few other implementations of `corfu--truncate-string-to-pixel-width`. The current implementation is as fast as it can get in Elisp. Any further improvement will have to involve a `string-glyph-width` implemented in C. I've attempted to implement it in Elisp with `char-displayable-p`, some undocumented functions in `composite.el` and a bunch of low level font functions. While the result was 2x faster than `string-pixel-width`, it was also wrong. A correct implementation will involve searching for the correct font from the faces in Elisp, which is extremely complex and would slow down the function by 2 orders of magnitude due to crazinesses such as anonymous faces and merging face attributes, and it still had to delegate to `string-pixel-width` for glyphs with display text properties.

That said, this PR's implementation still returns in < 2s for this benchmark emulating a `corfu-count` of 50 with default `corfu-max-width`.

```elisp
(let ((cands (cl-loop repeat 50 collect (list "candidatecandidatecandidat👱‍♀️🤬💁‍♂️も龢candidatecandidatecandidat👱‍♀️🤬💁‍♂️" "prefix" "suffixsuffixsuffixsuffixsuffixsuffixsuffixsuffix")))) (benchmark-run 60 (corfu--format-candidates cands 0)))
```

I think 30 FPS for extreme cases is quite alright.

jdtsmith:
> The result is the final width calculation, which is in pixels, is off by the difference of the number of pixels between the icon image widths and the default font widths multiplied by the number of glyphs in these different widths, resulting in the annotation being cut of

This by the way is why `kind-icon` explicitly creates icons exactly two `frame-char-width`s wide, to make this difference zero.

wyuenho:
What also means your icons are now rectangles, with uncustomizable and noticable gaps on both ends that don't fit into the rest of the text. You shouldn't have to do that if Corfu can align things in pixels.

jdtsmith:
They are SVGs that fill the 2 chars without gaps. 

But I'm definitely in favor of pixel alignment; @minad and I looked into it early in corfu's development and concluded it was too slow. 

One question I had for you both: can child frames have margins?  If so, you could put the scroll bar in the right margin and use specified space alignment relative to that.

wyuenho:
> They are SVGs that fill the 2 chars without gaps.

By gap, I mean empty space, AKA the margins left and right to the SVGs. They are quite noticeable in your screenshots and don't fit into the kerning of the rest of the text.

> One question I had for you both: can child frames have margins? If so, you could put the scroll bar in the right margin and use specified space alignment relative to that.

This is similar to what was suggested in emacs-devel. Frames have no margins, but the root window can. Again, the problem is calculating the width required for the child frame. At the time of such calculation, `string-pixel-width` uses the current window's width to calculate the relative space alignment offset, where in fact you want to align to the width of the child frame's window, of which it is unknown at this point in time. Eli suggested just blitzing the strings into the child frame's window, readjust the paddings afterwards, and then refit the frame to the new width. There are various ways I can think of to do this but they all sound insane to me and will also result in similar to what the Web frontend people called sudden [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content.).

> I looked into it early in corfu's development and concluded it was too slow

How slow is slow?

jdtsmith:
> By gap, I mean empty space, AKA the margins left and right to the SVGs. They are quite noticeable in your screenshots and don't fit into the kerning of the rest of the text.

Oh I see.  The icons are not text, and the "gaps" are padding put there _by design_.  It would look very awkward IMO to have no padding/normal text spacing, especially when the background color of the icon is styled. 

> Frames have no margins, but the root window can.

Yes naturally I meant windows in child-frames.  Since we are limited to `corfu-max-width`, would it not suffice to estimate the content width as we do now, clipping the child frame to the max?  Worst case scenario is slightly more/less space than would be ideal.  I deal with a similar issue with right-alignment in the mode-line of `mlscroll`, and this approach works well enough.

> The version that was in compat is 6x as slow. 

I see, so maybe its less of an issue now.  I recall it was slow in the sense that it would dominate the corfu popup display time.  Since some people have very slow machines/graphics, it's generally best for display itself not to dominate.

> There are various ways I can think of to do this but they all sound insane to me and will also result in similar to what the Web frontend people called sudden [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content.).

Not if they occur in one redisplay cycle.  Have you tried this approach?  In favor of it working out: `corfu` already redisplays the child frame with altered width as you scroll (rapidly) through candidates.

wyuenho:
> Oh I see. The icons are not text, and the "gaps" are padding put there _by design_. It would look very awkward IMO to have no padding/normal text spacing, especially when the background color of the icon is styled.

Yes, but the margins on your icons are not normal text spacing. You just took the default font width and divided it by 2. Text size icons should be treated like text in that it should respect the surrounding text's kerning. It's ok that they are wider, as they are images, but it's not ok that they look they don't belong to the same line. The prefix icons are ancillary information, they shouldn't scream out at you and say hey look at me, I'm way over here, separated from the rest. To your credit, your default icon choices make a lot of sense, so they are probably worth looking at more than the VSCode icons above.

> Yes naturally I meant windows in child-frames. Since we are limited to `corfu-max-width`, would it not suffice to estimate the content width as we do now, clipping the child frame to the max? Worst case scenario is slightly more/less space than would be ideal. I deal with a similar issue with right-alignment in the mode-line of `mlscroll`, and this approach works well enough.
> 

Did you just propose leaving 5 out of 6 problems solved in this PR unsolved?

> Not if they occur in one redisplay cycle. Have you tried this approach? In favor of it working out: `corfu` already redisplays the child frame with altered width as you scroll (rapidly) through candidates.

That's why there's a `corfu-min-width` to reduce this jarring effect. Thanks for remaining me the `inhibit-redisplay` hammer in `corfu--popup-show`.

I'm experimenting with deleting all truncation code and just let Emacs `truncate-line` do its thing. `string-pixel-width` calls are "batched" now by abusing its internal implementation that I'm not supposed to know. The right fringe is modified to show a bitmap ellipsis. The remaining trouble is the right margin. I want to show the scroll bar in it, but the smallest width a margin can be is 1 column. This means the scroll bar is going to be a fat boy but hopefully the tradeoff for performance is worth it.

wyuenho:
@minad @jdtsmith Truncation is now left to the window. RIP beautiful thin scroll bar. It's now rendered on the right margin, so min width of it is 1 column. `string-pixel-width` calls are now batched so every time `corfu--popup-show` is called, there are only 5 + `corfu-count` called instead of 100s previously. The [benchmark here](https://github.com/minad/corfu/pull/508#discussion_r1807943338) shows `corfu--format-candidates` returns in 0.1s, which is competitive with the current implements in HEAD.

Screenshot 2024-10-28 at 12 37 45 AM
Screenshot 2024-10-28 at 12 37 52 AM
Screenshot 2024-10-28 at 12 38 49 AM
Screenshot 2024-10-28 at 12 41 54 AM
Screenshot 2024-10-28 at 12 44 12 AM
Screenshot 2024-10-28 at 12 44 59 AM
Screenshot 2024-10-28 at 12 45 24 AM
Screenshot 2024-10-28 at 12 46 29 AM

corfu.el
wyuenho: Ok after a couple more days of testing, I think you are right, `string-pixel-width` can be put back into compat.

The reason it returns 0 in https://github.com/emacs-dashboard/emacs-dashboard/issues/430 is likely because the current buffer was on a 0 width window. That Elisp shim is 6x slower in my testing, but it's also faithful to the implementation in Emacs 29, it even has the same bugs which leads to https://github.com/seagle0128/doom-modeline/issues/601. There's nothing you can do about the mode line bug because you can't do [this](https://github.com/emacs-mirror/emacs/blob/emacs-30/src/xdisp.c#L11905-L11908) in Elisp, but it's equivalent with Emacs 29's implementation. The `buffer-text-pixel-size` implementation in https://github.com/emacs-compat/compat/commit/7c17c96c1e0e98033481394e05b20fc3399a373c is correct, you just need to copy https://github.com/emacs-mirror/emacs/blob/emacs-29/lisp/emacs-lisp/subr-x.el#L318-L333

jdtsmith:
Haven't looked in any detail here, but could you place all formatted text (prefix, candidate, space, suffix) in a hidden buffer without any `:align`, switch to it in the current window, measure the full width of the formatted "page" of data with `window-text-pixel-size`, and see if it is wider than `corfu-max-width` (in pixels)? 

If not, you have no overflow, so proceed in confidence.  

If so, go back and calculate pixel-width line by line, truncating each candidate, if needed, to allow the suffix to remain visible once right-aligned.  Here the idea is that only rare candidates will trigger this slower path.

minad:
@wyuenho 

> RIP beautiful thin scroll bar.

Would it work to use the fringe for the scroll bar? The fringe width is pixel precise while the margin is not. Losing the thin scroll bar is not an option, while I don't think that we absolutely need the ellipsis to be shown for truncation.

@jdtsmith 

> If so, go back and calculate pixel-width line by line, truncating each candidate, if needed, to allow the suffix to remain visible once right-aligned. Here the idea is that only rare candidates will trigger this slower path.

I prefer if there is no slow path at all. To be clear, I am happy with the status quo. With icon libraries like yours which properly use fixed columns, there is no alignment problem in most cases as long as fixed pitch fonts are used.

wyuenho:
> Would it work to use the fringe for the scroll bar? The fringe width is pixel precise while the margin is not. Losing the thin scroll bar is not an option, while I don't think that we absolutely need the ellipsis to be shown for truncation.

Possibly, the ellipsis are actually *NOT* using the fringe but actually using something called [display table](https://www.gnu.org/software/emacs/manual/html_node/elisp/Display-Tables.html#:~:text=A%20display%20table%20is%20a,to%20a%20display%20table%20object.) to render now. We may be able to have both, I'll look into it this week.

wyuenho:
I can't see your picture, you need to upload again.

wyuenho:
Thin scroll bar resurrected. Since it now uses the fringe, the ellipsis can't be shown, which is the current behavior. I've added an option to use the margin to show the scroll bar, which will have ellipsis for truncation. Default is the current behavior, thin scroll bar and no ellipsis.

jdtsmith:
> I can't see your picture, you need to upload again.

This was just your picture from above after max-width=55.

wyuenho:
I don't know understand what you mean. Every line has their padding and annotations hidden. Why wouldn't they need truncation? If you are saying the ellipsis look busier than the screenshots in the description, then it's because all the custom truncation code is gone, which means I can't just delete segments and selectively truncated just the longest line. I can put some of that back to replicate this effect, but then I'm sure someone else will come along and ask, how come every line has their annotations hidden but only one line showing and ellipsis?

minad:
@wyuenho 

> Thin scroll bar resurrected. Since it now uses the fringe, the ellipsis can't be shown, which is the current behavior. I've added an option to use the margin to show the scroll bar, which will have ellipsis for truncation. Default is the current behavior, thin scroll bar and no ellipsis.

Thanks. This is good news. Please remove the ellipsis code and the user option since I want to keep the complexity low. Also I agree with @jdtsmith that the screenshots look busy with the ellipsis and Corfu has done well without it so far.

wyuenho:
> I mean that every line except one in that image could have had their right annotation (type) right aligned without colliding with the candidate. Only one candidate line truly needed truncation.

You mean this is acceptable???
Screenshot 2024-10-19 at 5 50 32 PM

If not, you are talking about giving every line a different padding not only respecting the longest of the string, but the frame size, i.e.

```
--------------------------------
prefix candidate    annotation |
prefix candidate    annotation |
prefix reallllllylongcandidate | annotation <- overflow, hidden by the frame width, AKA truncated
prefix candidate    annotation |
--------------------------------
```

Which means, when I resize the popup frame from the right, I'll see just the one overflow annotation. In what world have you seen what is essentially a table organized like this?

> 4. Perfect alignment in the face of over-wide content therefore requires pixel width measurement of all 3 components prior to popup display. Which is too expensive/complex.

Not anymore. I don't think you and I are on the same page. Please look at the changes first.

> So at least one of these has to "give" to improve the situation. Corfu currently assumes 1&2 are non-issues (and 3?). It is usually correct about this for most setups.

I've played around over the weekend and found a solution. None of these have to give, they have all been achieved with excellent performance.

> In a perfect world we could truncate the CANDIDATE and right-align SUFFIX using specified space with `:align-to`. But that is expensive and CANDIDATE is not trivially truncate-able "by hand".

This has already been achieved and there's no need to truncate "by hand". Please look at the changes first.

wyuenho:
BTW, I think I need to make a case for ellipsis. I've never really made one, I just assumed it'll be obvious, but obviously it's not obvious in hindsight. Let me come back with some examples later.

jdtsmith:
Currently corfu handles over-width _OK_.  When the content fits, it right-aligns everything (here I've set `max-width=24`):

image

And when it doesn't fit, it still preserves much of the SUFFIXes where it can :

image

I'd prefer those SUFFIXes which are visible to be right-aligned where possible, but at least the information is still (usually) visible.  I do note corfu sometimes reverts to "right-aligned but truncated" SUFFIXes, though I'm not sure why (probably a bug in the logic which determines whether right-alignment for all current candidates is possible):

image

To me, any new solution here should at least preserve and improve on the current situation in the default and typical case of PREFIX, CANDIDATE, and SUFFIX all being column-aligned.  The current version of this PR doesn't really preserve this simple case (though you may have other fixes in the queue):

image

Current corfu is also certainly somewhat faster when scrolling through.

jdtsmith:
> > I mean that every line except one in that image could have had their right annotation (type) right aligned without colliding with the candidate. Only one candidate line truly needed truncation.
> 
> You mean this is acceptable??? 

No it's not ideal, see my other message.   But it's preferable to losing all SUFFIX information because one line has gone rogue. 

> If not, you are talking about giving every line a different padding not only respecting the longest of the string, but the frame size, i.e.
> 
> ```
> --------------------------------
> prefix candidate    annotation |
> prefix candidate    annotation |
> prefix reallllllylongcandidate | annotation <- overflow, hidden by the frame width, AKA truncated
> prefix candidate    annotation |
> --------------------------------
> ```
> 
> Which means, when I resize the popup frame from the right, I'll see just the one overflow annotation. In what world have you seen what is essentially a table organized like this?

In this world I have certainly seen data organized like this.  

 ```
 ---------------------------------
 prefix candidate    annotation  |
 prefix reallllllylongcandidate… | annotation <- overflow truncated
 prefix candidate    annotation  |
 ---------------------------------
 ```

though one might instead prefer:

```
 ----------------------------------
 prefix candidate      annotation |
 prefix reallllllylon… annotation |
 prefix candidate      annotation |
 ----------------------------------
 ```

> > 4. Perfect alignment in the face of over-wide content therefore requires pixel width measurement of all 3 components prior to popup display. Which is too expensive/complex.
> 
> Not anymore. I don't think you and I are on the same page. Please look at the changes first.

Important to know: in this numbered list I'm stating the original problem as I understand it, not opining on your current solution.  I can't tell whether you agree or disagree with my assessment of the original problem.

> > So at least one of these has to "give" to improve the situation. Corfu currently assumes 1&2 are non-issues (and 3?). It is usually correct about this for most setups.
> 
> I've played around over the weekend and found a solution. None of these have to give, they have all been achieved with excellent performance.

In my current testing, I have experienced some modest slowdown compared to stock corfu.  The bigger issue is that SUFFIX data go missing.  See my other message. 

> > In a perfect world we could truncate the CANDIDATE and right-align SUFFIX using specified space with `:align-to`. But that is expensive and CANDIDATE is not trivially truncate-able "by hand".
> This has already been achieved and there's no need to truncate "by hand". Please look at the changes first.

This has not been achieved that I've seen.  Truncate _CANDIDATE_, preserve SUFFIX is my "perfect world" scenario.  Others may disagree.

I'm late to this conversation so you and @minad may feel like you're already converging, in which case I'll leave you to it.  I'd like to continue engaging to see if we can guide this towards a solutions which preserves and builds on corfu's _current strengths_ and improves the situation for non-fixed data, but speaking honestly, your quick-take reactions are making it more difficult.

**Comment by minad** [comment]
@wyuenho 

> BTW, I think I need to make a case for ellipsis. I've never really made one, I just assumed it'll be obvious, but obviously it's not obvious in hindsight. Let me come back with some examples later.

I see your point that an ellipsis would be nice to have, but having the two variants ellipsis+thick scroll bar and no ellipsis+thin scrollbar is not an acceptable solution with a customization option. I can elaborate on the reasons. In contrast, if we have ellipsis and scrollbar thickness as separate and disentangled options, I am happy with the addition, but this doesn't seem to be on the table given the technical limitations. Technically I prefer if we only use the fringe, since the margin has these problems with redisplay anyway.

File: corfu.el [hunk 5]
```diff
@@ -230,6 +240,27 @@ See also the settings `corfu-auto-delay', `corfu-auto-prefix' and
...
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   #b11111111111111
+   )
+  255 16 '(center . t))
```
minad, line 262: Why do we need a bitmap? Can we not simply use the entire fringe as scrollbar and simply adjust the background accordingly (scroll bar background, current background and default background=?

minad:
@jdtsmith

> No it's not ideal, see my other message. But it's preferable to losing all SUFFIX information because one line has gone rogue.

Agree. How does the new behavior compare to the current behavior?

> In my current testing, I have experienced some modest slowdown compared to stock corfu. The bigger issue is that SUFFIX data go missing. See my other message.

I want to state this very clearly - if there is any noticeable slow down, I will not accept the change. If we cannot get good performance with pixel alignment, so be it. However I will accept a measurable slow down as long as it is roughly in the same ball park.

> I'd like to continue engaging to see if we can guide this towards a solutions which preserves and builds on corfu's current strengths and improves the situation for non-fixed data, but speaking honestly, your quick-take reactions are making it more difficult.

Please stay involved here. The situation is not finally converging. I also find it difficult to discuss given the quick reactions. For example I pointed out the potential of a slow down early on, which was first brushed aside. Then it turned out that the slowdown of reformatting was factor 20 (!) in the first draft.

corfu.el
wyuenho: Nope. Because with emojis and variable fonts, even the lines are not in uniform heights. Is there any performance concern on this one?

wyuenho:
@jdtsmith 

> Currently corfu handles over-width _OK_. When the content fits, it right-aligns everything (here I've set `max-width=24`):
> 

So does this PR.

> And when it doesn't fit, it still preserves much of the SUFFIXes where it can :
> 
> image

I consider this a bug. This is an artifact of [this](https://github.com/minad/corfu/blob/6f5518ea58498af9cf86b8fafbdf1ef4d0997dcc/corfu.el#L737-L743), which progressively subtracts from the column widths, but when a line is shorter than the longest line, the `truncate-string-to-width` call in the loop does not add back enough padding to fit. The non-alignment is extremely distracting. This seems to be a bug that's been around long enough that users have treated it as a feature.

> I'd prefer those SUFFIXes which are visible to be right-aligned where possible, but at least the information is still (usually) visible. I do note corfu sometimes reverts to "right-aligned but truncated" SUFFIXes, though I'm not sure why (probably a bug in the logic which determines whether right-alignment for all current candidates is possible):

In this PR, it is possible to right align every line except the one that has to overflow (without ellipsis), which I assume you'll consider that an improvement, but it'll be odd when I resize the popup frame and see every line mis-aligned again. I can implement this, but I'm not convinced it's an improvement yet. I'll have to think about this one. Regardless, if your `corfu-max-width` happens to land right on where the candidates end. You'll still see all the suffixes truncated. This is also the current behavior.

As to the "right-aligned but truncated", that's because of a combination of all the annotations having the same width (they are all Method in your case) and `corfu-max-width` being set too low but not low enough to start truncating off the candidates.

> To me, any new solution here should at least preserve and improve on the current situation in the default and typical case of PREFIX, CANDIDATE, and SUFFIX all being column-aligned. The current version of this PR doesn't really preserve this simple case (though you may have other fixes in the queue):
> 
> image

Please show me a picture of this PR not aligning the columns.

> Current corfu is also certainly somewhat faster when scrolling through.

The current state of this PR has a `corfu--format-candidate` that can format strings in 360 FPS on a 4 year old computer, even if the 2 extra calls of `string-pixel-width` in `corfu--popup-show` slow things down, they are in the <0.01ms range. Are you sure you can feel the slowdown? Can you quantify this please?

> In this world I have certainly seen data organized like this.

```
---------------------------------
prefix candidate    annotation  |
prefix reallllllylongcandidate… | annotation <- overflow truncated
prefix candidate    annotation  |
---------------------------------
```

Which is the slow option in my original implementation, which we have ruled out due to performance concern. This will require "by hand" truncation, which is slow or imprecise.

> though one might instead prefer:

```
 ----------------------------------
 prefix candidate      annotation |
 prefix reallllllylon… annotation |
 prefix candidate      annotation |
 ----------------------------------
 ```
 
This would be ideal, but would require the user specifying the max width per column, à la Excel. I haven't seen an issue in this repo asking for it, but if they have, it'll likely be in conflict with performance and/or pixel-preciseness with "by hand"  truncation. This is a limitation of Emacs I'm afraid.

@minad  since the margin has these problems with redisplay anyway.

Is this regarding [this](https://github.com/minad/corfu/pull/508/files#diff-61e35fd9a4b9c1efde05e06d53a1461a486e7e923666faf86c8c59ffd936743dR537-R543) and [this](https://github.com/minad/corfu/pull/508/files#diff-de7cae98d81385e6cc0da9d17f5b27e5e4e80f01e4229f8d72e9eb8b87a4cf3cL384-L390)?

I apologize for the quick reactions. For @minad, there's a lot of missing context that I had to dig very hard to understand what you mean. It would be very helpful if you could provide them as I don't go through the emacs-devel mailing list or all your old issues on a daily basis. For @jdtsmith I understand this thread is becoming difficult to follow, but I think it would make it easier if you ask questions first if you are not sure about something. I apologize regardless.

**Comment by minad** [comment]
@wyuenho

> The current state of this PR has a corfu--format-candidate that can format strings in 360 FPS on a 4 year old computer, even if the 2 extra calls of string-pixel-width in corfu--popup-show slow things down, they are in the <0.01ms range. Are you sure you can feel the slowdown? Can you quantify this please?

I think my machine right now is around eight or ten years old but feels pretty fast with my full Emacs setup. If I can feel a slowdown I will let you know. I am not extremely sensitive to this, but just to give you some perspective, I've had complaints about Corfu latency issues in the millisecond range, which I also found hard to see or consider relevant. Yet people have sent me patches improving these issues, which I've merged.

> For @minad, there's a lot of missing context that I had to dig very hard to understand what you mean. It would be very helpful if you could provide them as I don't go through the emacs-devel mailing list or all your old issues on a daily basis.

I try my best, but I cannot spent much time on this. Ideally you don't simply brush aside comments before seeing the hard facts, like in the benchmark discussion. Also regarding alignment, it shouldn't be hard to compare the status quo with your PR side by side, in particular if @jdtsmith is also helping here. Note that pixel perfect alignment is not a goal we will pursue despite all costs. It must be a reasonable trade-off - this means only moderate or no slow down, and ideally no regressions regarding the formatting in scenarios where Corfu works well right now. The comments regarding the kind-icon formatting from you were also quite unnecessary - kind-icon was deliberately designed like this to fit the fixed width requirements. In other words kind-icon is the only package compliant with the Corfu margin formatter API as it stands now. I also find it quite beautiful and certainly not a typographical disaster.

> For @jdtsmith I understand this thread is becoming difficult to follow, but I think it would make it easier if you ask questions first if you are not sure about something. I apologize regardless.

I don't think it is hard for him. JD has been involved in Corfu for quite a while and I had discussed a lot of details with him, regarding performance, regarding formatting, regarding completion behavior and so on.

File: corfu.el (hunk 5 above)
Comment (minad) [reply corfu.el:262]: I do not understand this - what does this have to do with the height? Can I not simply configure the background color of the fringe and adjust the fringe width to a fixed pixel width? Then the fringe alone will function as the scroll bar. Or do I always need a bitmap? If yes, why?

**Comment by wyuenho** [comment]
>  I've had complaints about Corfu latency issues in the millisecond range, which I also found hard to see or consider relevant. Yet people have sent me patches improving these issues, which I've merged.

I see why you are getting slightly burnt out :) Such is the state of affairs for Emacs.

> regarding alignment, it shouldn't be hard to compare the status quo with your PR side by side

Yep, and I'll wait for his screenshots. I haven't seen any of this PR yet. That jagged case tho, I'll consider right aligning it, with an option to select the behavior of the state of this PR as of now, defaulting to right aligning but truncating the few overflows. Let me know if this is agreeable to both of you.

BTW, please do not be fooled by the following picture. I set `corfu-min-width` to 50 to reduce the popup frame from flickering left and right when scrolling.

Screenshot 2024-10-28 at 12 41 54 AM

> kind-icon was deliberately designed like this to fit the fixed width requirements. In other words kind-icon is the only package compliant with the Corfu margin formatter API as it stands now. I also find it quite beautiful and certainly not a typographical disaster.

To be clear, I'm not saying kind-icons is a typographical disaster, far from it, the icon choices are really well done and far superior than VSCode's boxes for methods and fields. It's just somewhere along the line background colors on icons were mentioned and I looked at the screenshot in the repo's README, all the negative space separations screamed out at me and I lost it. Normally I assume, you don't use kind icons that way, as seen in @jdtsmith 's screenshots in this thread.

Anyway, I agree pixel alignment shouldn't be a goal to pursue in all costs, as with pretty much most engineering decisions in life.

**Comment by minad** [comment]
> Anyway, I agree pixel alignment shouldn't be a goal to pursue in all costs, as with pretty much most engineering decisions in life. Apologies for being a little frustrated as I've been working on this PR for weeks now.

I understand this. But note that we've had also already looked into pixel alignment and didn't pursue it early on, because of performance, missing APIs and so on. So it is a difficult problem and it could happen that we may not succeed. But this is a limitation Emacs then and not due to your lack of trying.

> Yep, and I'll wait for his screenshots. I haven't seen any of his from this PR yet. That jagged case tho, I'll consider right aligning it, with an option to select the behavior of the state of this PR as of now, defaulting to right aligning but truncating the few overflows. Let me know if this is agreeable to both of you.

Generally, if the status quo is preserved in cases where Corfu works well, it is acceptable. Then we have a pure improvement and no regressions. If there are edge cases where the new behavior will be worse and we add customization options to revert back to the old behavior, the answer tends to be that this is not acceptable. However we may accept minor regressions if the overall trade-off seems worth it (without a customization option). There were proposals where I don't see that, e.g., the thick scroll bar, where we would end up with two solutions, but both a little bit flawed.

**Comment by wyuenho** [comment]
> Generally, if the status quo is preserved in cases where Corfu works well, it is acceptable. Then we have a pure improvement and no regressions. If there are relevant cases where the new behavior will be worse and we add customization options to revert back to the old behavior, the answer tends to be that this is not acceptable. However we may accept minor regressions (in edge cases) if the overall trade-off seems worth it (without a customization option).

I need this to be clear. I'm proposing improving on the following situation by right aligning and showing the full suffixes when we can, but overflow the few we can't. I'm proposing this to be the default behavior so as to "improve" on the the current situation, which I'm still not sure whether both of you consider "working well". Since I do not consider this to be working well, as I do not set such low `corfu-min-width`, I propose adding an option to just truncate every line uniformly similar to what's shown [here](https://github.com/minad/corfu/pull/508#issuecomment-2440275817), but without ellipsis and thick scroll bar.

image

> There were proposals here, where I didn't see that, e.g., the thick scroll bar with ellipsis and a customization option, where we would end up with two solutions, but both a little bit flawed.

Just out of curiosity, it was never made clear to me, why is the thin scroll bar nonnegotiable? autocomplete, company and company-box all have had thick scroll bar for many many years and I've never heard a complain. The thin scroll bar with adjustable width is beautiful for sure, but is it *essential*?

**Comment by minad** [comment]
> I need this to be clear. I'm proposing improving on the following situation by right aligning and showing the full suffixes when we can, but overflow the few we can't.

I am not exactly sure what you are proposing. Can you provide side by side screenshots of the status quo and your version. I am a bit worried about the turn this is taking. We are not talking about strict improvements, we are talking about trading one thing for something else. It is unfortunate that we have to make such trade-offs, but unfortunately the Emacs display system seems not (yet) good enough to give us sufficient flexibility. To be clear about this - the reason why this is taking so long is that we have to make these trade-offs. The situation would be very different and much simpler if this were a strict improvement (without performance problems), as we initially assumed when you proposed this.

> Just out of curiosity, it was never made clear to me, why is the thin scroll bar nonnegotiable? autocomplete, company and company-box all have had thick scroll bar for many many years and I've never heard a complain. The thin scroll bar with adjustable width is beautiful for sure, but is it essential?

Every change must be justified, and the status quo works well enough for me and many users. The thin scrollbar is a relatively unobtrusive indicator and doesn't waste much space. Company also introduced a thinner scrollbar if I recall correctly, but it will still require a full character for the whole margin due to the implementation via overlays. This means Company shows some more padding between the scrollbar and the text. So this design generally seems to be appreciated. Note that the current scrollbar can be adjusted in width, so it is not necessarily thin. If I understand correctly, this would not be possible anymore with the scrollbar in the margin, which quite obviously would be a regression in behavior. 

The simple answer is that I am not willing to sacrifice the thin (adjustable) scrollbar to achieve pixel precise alignment, given that pixel precise alignment is not a problem in many cases. Alignment currently works even with icons (kind-icon). Problems 1-3 are actually bugs in nerd-icons-corfu, since nerd-icons-corfu could also use precisely two characters or squeeze the icon into a single character.

File: corfu.el
Comment (wyuenho) [inline corfu.el:1]: @jdtsmith This should workaround that Emacs bug in your [last picture in this comment](https://github.com/minad/corfu/pull/508#issuecomment-2444935615) by doing a more precise right padding calculation.

Comment (wyuenho) [inline corfu.el:1]: @jdtsmith This commit should workaround that Emacs bug in your [last picture in this comment](https://github.com/minad/corfu/pull/508#issuecomment-2444935615) by doing a more precise right padding calculation instead of relying on the window to truncate.

File: corfu.el (hunk 5 above)
Comment (wyuenho) [reply corfu.el:262]: > Can I not simply configure the background color of the fringe and adjust the fringe width to a fixed pixel width? Then the fringe alone will function as the scroll bar. Or do I always need a bitmap? If yes, why?

The fringe is not just what you specify as text properties as the fringe, but an actual area on 2 sides of an [entire window](https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Windows.html). If you just put a face on the fringe, the entire fringe on the side of a buffer looks like one solid scroll bar. As far as I could tell, you [always need a bitmap](https://www.gnu.org/software/emacs/manual/html_node/elisp/Fringe-Bitmaps.html), it can't be a plain string or anything else, tho it need not to be in an overlay as suggested in the documentation. This makes `corfu--popup-show` even faster as we don't have to scan some buffer positions to decide where to put the overlay.

You can only have a max *fringe* width of 16 pixels. The *bitmap* tho, has a max height of 255 pixels, which is of course not documented but have to be found by reading the [C source](https://github.com/emacs-mirror/emacs/blob/75ca15f2a3146b17fde5e2a50ad9102c6108ba7f/src/fringe.c#L1573). When I was testing, I started with `filled-rectangle`, that didn't work, the top and bottom of the bitmap had the background poking through, so the scroll bar looked like a dashed line. Then I changed to this definition with a filled 16x16 sqaure, but without `'(center . t)`, just `'center`, that worked, until I resized the default font height to 18pts, then it starts to look like an I in a serif font clipped from the middle. That's when I put in `'(center . t)`. This worked, so I stopped testing. I just spent 15mins during my lunch break on this.

> Could one also use an empty bitmap, all zero or all one of size 1x1 with some cyclic repetition?

This is an idea I had not thought about, but it works! I can delete the foreground color too.

**Comment by jdtsmith** [comment]
> For @jdtsmith I understand this thread is becoming difficult to follow, but I think it would make it easier if you ask questions first if you are not sure about something. I apologize regardless.

Thanks.  Maybe you missed it as it came late, but I did ask a question: 

> Does that capture the issues as you see them?

I.e., rather than iterating on a technical solution to a set of problems as seen from one person's perspective, sometimes I find it's very useful to zoom out and identify the issues and constraints more generally.  I.e. first _agree on the problem(s)_.

**Comment by jdtsmith** [comment]
Some comparison screenshots with `max-width=24`.  This is fixed width text with fixed 2-char icons.  

## Aligned but truncated SUFFIX

| MAIN | PR |
|-- | -- |
image  | image

## Staggered SUFFIX

| MAIN | PR |
|-- | -- |
| image | image |

I'd say neither of these is ideal.  Clearly main has some strange trigger which causes a flip between aligned-but-truncated vs. staggered SUFFIX, which may indeed be a bug.  But it does preserve _more_ SUFFIX information.  I also note the continuation chars in the fringe.

In my view, in the perfect world, in both of these cases, SUFFIX would _right-align_ on each line where it has the space to do so, and where it cannot, either the SUFFIX gets lopped off (with display-based truncation) or (even better, but harder) room is made with candidate truncation.

I.e., something like this mock-up (made by editing the above):

image

Once this is sorted it can be considered whether this is also possible with pixel-precise placement for some subset of PREFIX/CANDIDATE/SUFFIX being non-fixed-char-aligned.

**Comment by minad** [comment]
@jdtsmith 

> I'd say neither of these is ideal. Clearly main has some strange trigger which causes a flip between aligned-but-truncated vs. staggered SUFFIX, which may indeed be a bug. But it does preserve more SUFFIX information. I also note the continuation chars in the fringe.

Honestly I don't recall why this happens. It may be a bug in the alignment code or the display engine. But I've never seen this as a big problem. Besides that the existing behavior in main is clearly better.

> In my view, in the perfect world, in both of these cases, SUFFIX would right-align on each line where it has the space to do so, and where it cannot, either the SUFFIX gets lopped off (with display-based truncation) or (even better, but harder) room is made with candidate truncation. I.e., something like this mock-up (made by editing the above):

Agree. This would be better.

@wyuenho

Regarding the scrollbar in your PR - it is different to the one in the main branch. There is a larger padding with background color between the scrollbar and the content. The current line indicator doesn't extend completely to the scrollbar and it doesn't extend completely to the right border.

**Comment by wyuenho** [comment]
@minad @jdtsmith I see that the truncation right arrow is still poking thru on the fringe, did f7c91b58229644599c710e9a1eadf2f2ecfeda53 still hadn't worked around the Emacs bug on Linux? Did you do a fresh install and restart Emacs and everything?

**Comment by wyuenho** [comment]
> Some comparison screenshots with `max-width=24`. This is fixed width text with fixed 2-char icons.
> 
> ## Aligned but truncated SUFFIX
> MAIN	PR
> ...snip...
> Once this is sorted it can be considered whether this is also possible with pixel-precise placement for some subset of PREFIX/CANDIDATE/SUFFIX being non-fixed-char-aligned.

I think I see why you think pixel column alignment has not been achieved haha. Let me illustrate with the following example.

Screenshot 2024-10-30 at 11 31 43 AM
Screenshot 2024-10-30 at 11 31 59 AM
Screenshot 2024-10-30 at 11 32 16 AM
Screenshot 2024-10-30 at 11 32 28 AM
Screenshot 2024-10-30 at 11 33 46 AM

Do you see the "problem" now? If you resize the popup frame from the right, do the clipped suffixes suddenly appear padded and aligned correctly to you?

File: corfu.el [hunk 6]
```diff
@@ -776,7 +776,7 @@ terminal."
              do
              ;; `corfu-current' may affect frame-width too
              (when (= i curr)
-               (cl-loop for s in (list cand prefix suffix)
+               (cl-loop for s in (list cand prefix suffix marginl marginr)
```
wyuenho, line 779: @minad This fixes the following issue

> The current line indicator doesn't extend completely to the scrollbar and it doesn't extend completely to the right border.

corfu.el
```diff
@@ -1084,8 +1091,9 @@ A scroll bar is displayed from LO to LO+BAR."
     (with-current-buffer (corfu--make-buffer " *corfu*")
       (let* ((ch (default-line-height))
              (cw (default-font-width))
-             (mr (ceiling (* cw corfu-right-margin-width)))
-             (bw (ceiling (* cw corfu-bar-width)))
+             (mr (max 0 (ceiling (* cw corfu-right-margin-width))))
+             (bw (max 0 (ceiling (* cw corfu-bar-width))))
+             (bw (min mr bw))
```
wyuenho, line 1077: @jdtsmith This commit should workaround that Emacs bug in your [last picture in this comment](https://github.com/minad/corfu/pull/508#issuecomment-2444935615) by doing a more precise right padding calculation instead of relying on the window to truncate.

minad:
I decided to close this PR since we are not making progress here and I also don't see this feature happening in the near future as long as Emacs doesn't provide better facilities. We can revisit this problem later on again, after the situation cooled down. 

There is too much debate and no real progress, and a lot of miscommunication. I continuously have the feeling that I have to defend the way Corfu works right now, while for new changes the default modus should be that the new change should not degrade existing functionality, nor should we trade them.

This comment was the final drop:
https://github.com/LuigiPiucco/nerd-icons-corfu/issues/13#issuecomment-2446648332

> Indeed, as predicted in my first response, but even if it's misaligned, it's still arguable whether it is a bug in downstream packages, especially if you've documented the requirement just now.

@wyuenho "especially if you've documented the requirement just now" - This does not change the fact that this requirement has already been in place. I don't recall why I haven't documented this. Probably it has been simply an oversight or I wanted to leave it open to icon package maintainers to ignore the alignment in favor of icon typography. In any case, @jdtsmith and I were aware of this fact a long time ago.

I am tired of the attitude and expectation that the pixel alignment feature has to simply make it in. I do not want to feel bad about my FOSS work, but this PR and your attitude again create such a feeling. In order to get your change in you proposed to sacrifice performance, you proposed to drop compatibility with supported Emacs versions, you proposed to sacrifice the existing styling and you proposed to give up on the existing alignment, which is mostly better as @jdtsmith demonstrated. These things are not going to happen.

You continuously question facts - I am all for being critical, but let's please look at the objective performance numbers. If I am telling you that things are how they for historical reason, this has to be acknowledged, even if it is an ugly reality. I am not lying about the requirement, it is a historical limitation in place because of the lack of pixel alignment. I am not making this up just now.

PR #509, opened 2024-10-21
Pixel perfect Corfu extension

wyuenho:
Separate #508 to an extension that requires Emacs 29

https://github.com/minad/corfu/pull/508/files#r1807971159

Alternatively, since the pixel mode switching logic is controlled by `corfu--pixel-wise` anyway, I can move the extension back into the core and turn `corfu--pixel-wise` to a `defcustom` that'll automatically set to t for Emacs >= 29.


minad:
Thanks, but see my https://github.com/minad/corfu/pull/508#discussion_r1807967173.

PR #521, opened 2024-11-01
Use fringe for the scroll bar

minad:
This ensures that the scroll bar is always visible, even if wide characters lead to truncation of the popup content. Before this change, the scroll bar could be pushed out of the window by the content.

This PR can be tested for example with cape-emoji where the emojis in the annotation are wider than the fixed width characters.

~The problem with this PR is that it breaks the terminal scrollbar. Therefore probably the old code has to be kept and only on GUI (fringe-bitmap-p function available) the fringe should be used.~ (EDIT: Fixed now)

cc @jdtsmith

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -479,7 +479,8 @@ FRAME is the existing frame."
     ;; overrides the parameter `tool-bar-lines' for every frame, including child
     ;; frames.  The child frame API is a pleasure to work with.  It is full of
     ;; lovely surprises.
-    (let* ((is (frame-parameters frame))
+    (let* ((win (frame-root-window frame))
+           (is (frame-parameters frame))
            (should `((background-color
                       . ,(face-attribute 'corfu-default :background nil 'default))
                      (font . ,(frame-parameter parent 'font))
@@ -488,9 +489,9 @@ FRAME is the existing frame."
                      ,@corfu--frame-parameters))
            (diff (cl-loop for p in should for (k . v) = p
                           unless (equal (alist-get k is) v) collect p)))
-      (when diff (modify-frame-parameters frame diff)))
-    (let ((win (frame-root-window frame)))
-      (unless (eq (window-buffer win) (current-buffer))
+      (when diff (modify-frame-parameters frame diff))
+      ;; XXX HACK: `set-window-buffer' must be called to force fringe update.
+      (when (or diff (eq (window-buffer win) (current-buffer)))
         (set-window-buffer win (current-buffer)))
       ;; Disallow selection of root window (gh:minad/corfu#63)
       (set-window-parameter win 'no-delete-other-windows t)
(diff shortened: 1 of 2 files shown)


jdtsmith:
Good idea.  For me the fringe is large and cuts off one suffix position:

image

minad:
> Update: I had set bar-width to 1.5 (though it wasn't doing anything, because corfu-right-margin-width was overriding it). Now looks like:

Alright. Thanks.

> I do notice a small gap between the text and bar, and we should probably support wider corfu-bar-width without truncation.

This gap intentionally has been there all along due to corfu-right-margin-width. Usually corfu-bar-width takes only a fraction of the margin space.

corfu.el
```diff
@@ -1021,17 +1023,18 @@ A scroll bar is displayed from LO to LO+BAR."
...
-                     (concat (propertize " " 'display `(space :align-to (- right (,mr))))
-                             (propertize " " 'display `(space :width (,(- mr bw))))
+             (sbar (if fringe
+                       #(" " 0 1 (display (right-fringe corfu--bar corfu-bar)))
+                     (concat (propertize " " 'display `(space :align-to (- right (,bw))))
                              (propertize " " 'face 'corfu-bar 'display `(space :width (,bw))))))
+             (cbar (and fringe #(" " 0 1 (display (right-fringe corfu--bar corfu-current)))))
              (pos (posn-x-y pos))
-             (width (+ (* width cw) ml mr))
+             (width (+ (* width cw) ml mr (if fringe (- bw) 0)))
```
jdtsmith, line 1037: I'm not sure we should remove `bw` when there is a fringe.  The bar used to "live inside" the right margin, but now it's in the right fringe which is shown in addition to the right margin.  The specified frame width does not include the fringe I think.

minad: Maybe there is a misunderstanding. If we have a scrollbar, we use the fringe, which is not part of the frame width. Therefore we have to subtract this additional width to keep the same overall frame width. Does this make sense?

For testing you can try to set `fringe` to nil:

https://github.com/minad/corfu/blob/190aca1d8556789c45f352ea44adf5b7a4c595c3/corfu.el#L1027

In both cases the resulting picture should look identical. The difference only occurs if we have non-fixed characters, like the emojis of cape-emoji.

minad: Not sure I understand. It is over the top of my head right now and I am bothering about too many things right now :/

Note that `corfu-bar-width` should always be less or equal to `corfu-right-margin-width`. Does the cutting off happen then too?

jdtsmith: The way I think of it: `corfu--format-candidates` asks for a `width` for _content_: prefix + candidates + suffix.  It does not care about margin and it does not care about the bar.   So if a large bar "takes up" area that `format-candidates` needed for its actual content, the content will no longer fit. 

But your way also makes sense, now with the new enforcement: bar always "eats" right-margin, not text content area.  This now works fine.  

I do notice that updating `corfu-bar-width` requires killing the ` *corfu*` hidden buffer to take effect.

minad: > I do notice that updating corfu-bar-width requires killing the  *corfu* hidden buffer to take effect.

I think this is fixed by https://github.com/minad/corfu/pull/521/commits/a1e494e29ac02a6c4fe9ffe6e1fe70910ecf1848.

PR #524, opened 2024-11-04
Avoid truncate-string-to-width

minad:
1. The first commit avoids `truncate-string-to-width` if the fringe is used for the scrollbar. The first commit can be checked separately to see how the result looks without the second step.
2. The second commit additionally moves the margin into the fringe too, such that the truncated text doesn't leak into the margin. I am not entirely happy with the way this is implemented.

EDIT: Turns out that fringe bitmaps have a width restriction, which makes 2. infeasible. So maybe use only 1. But there the margin is effectively lost.

EDIT2: The result works reasonably well now. The fringe/margin is restricted to 16 pixel. The existing configuration variables and faces are also preserved as is.

cc @jdtsmith

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -5,13 +5,9 @@
 * Development
 
 - Require Emacs 28.1.
-- Use fringe to display scroll bar. Make sure that the scroll bar cannot be
-  pushed outside the child frame by the content. This affects for example
-  ~cape-emoji~.
-- New face ~corfu-scroll-bar~ which replaces the old face ~corfu-bar~. For technical
-  reasons the foreground color is used for the scroll bar, and not the
-  background as for the old face.
-- Renamed ~corfu-bar-width~ to ~corfu-scroll-bar-width~ for consistency.
+- Use fringe to display scroll bar. This change improves performance and makes
+  sure that the scroll bar cannot be pushed outside the child frame by the
+  content. This affects for example ~cape-emoji~.
 - Improve suffix alignment.
 
 * Version 1.5 (2024-07-26)
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -8,6 +8,10 @@
 - Use fringe to display scroll bar. Make sure that the scroll bar cannot be
   pushed outside the child frame by the content. This affects for example
   ~cape-emoji~.
+- New face ~corfu-scroll-bar~ which replaces the old face ~corfu-bar~. For technical
+  reasons the foreground color is used for the scroll bar, and not the
+  background as for the old face.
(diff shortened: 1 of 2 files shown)


jdtsmith:
Thanks, this works well here and seems (subjectively) even snappier.  I worry a little bit about the restricting margin to 16 pixels, since some HDPi Linux systems report high pixel density chars, e.g. 20x30.  Mac HDPi simply lies about pixel sizes (by 2x) and then just shows extra pixels for images/glyphs/etc.  Pretty clever actually.

I personally wouldn't mind over-long candidates "running into the bar" as it would be an indication that they've been truncated (without taking up space with ellipses).  But this method should be fine; I doubt the default settings would ever cause problems on HDPi systems.

BTW, one point made over on the other site is that it can be jarring to have the window change width so much.  I tend to agree, and 15->100 seems like a pretty large range.

minad:
>  I worry a little bit about the restricting margin to 16 pixels, since some HDPi Linux systems report high pixel density chars, e.g. 20x30.

I worry about that too. However the default is `corfu-right-margin-width 0.4` (8 pixel for a char width of 20). Only if you increase it to 0.8 you will hit the limit. However the problem becomes more pronounced when you use large font sizes.

Ultimately Emacs has to adopt larger fringe bitmaps I think. I made this report and absurdly one of the Emacs devs responded that they actually want smaller fringe bitmaps (4 pixel)
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=59243 🫤 

> I personally wouldn't mind over-long candidates "running into the bar" as it would be an indication that they've been truncated (without taking up space with ellipses).

It doesn't look good. You can try the first commit only. To be clear, the goal is to avoid programmatic truncation altogether which will help with pixel alignment at some later point. And it also helps performance.

> BTW, one point made over on the other site is that it can be jarring to have the window change width so much. I tend to agree, and 15->100 seems like a pretty large range.

I agree. The problem is that the defaults have to work for many different major modes and small popups work rather well in Eshell for example. In programming major modes the default should maybe be 40 to 80? It should be possible to adjust these settings buffer-locally.

wyuenho made many good points, and there are multiple interesting improvements over there. I realized that the source of frustration was the different pace of work speed. I am iterating painfully slowly here, while he wanted to move faster and experiment with other new features. I also do that on a prototyping stage and then it is of course perfectly fine to not bother about compatibility and minor regressions.

minad:
@jdtsmith See https://github.com/minad/corfu/commit/37d4ed3ca0a7c314382af64bda643a9dbf31c469 regarding the popup width fluctuations.

jdtsmith:
Looking good here.

PR #533, opened 2024-12-08
Override faces from completion styles with our own in preview

wyuenho:
Some completion styles like corfu-prescient will apply faces to the candidate strings deep inside the completion-in-region machinery before corfu gets to them. While these faces are useful in the completion popup, they aren't so useful in the preview. This is evident in the first screenshot in the [README](https://github.com/minad/corfu/raw/screenshots/light.png?raw=true), here's another example in TypeScript where they make even less sense.

ScreenRecording2024-12-08at5 49 27PM-ezgif com-optipng

This PR makes a copy of the candidate string and overrides the face property with a new face that by default inherits from corfu-default, so the user can customize the face if they so choose.
ScreenRecording2024-12-08at5 58 11PM-ezgif com-optipng

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -210,6 +210,10 @@ settings `corfu-auto-delay', `corfu-auto-prefix' and
     (t :background "gray"))
   "Default face, foreground and background colors used for the popup.")
 
+(defface corfu-preview
+  '((t :inherit corfu-default))
+  "Face used for the candidate preview.")
+
 (defface corfu-current
   '((((class color) (min-colors 88) (background dark))
      :background "#00415e" :foreground "white" :extend t)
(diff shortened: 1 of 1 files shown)


minad:
This problem does neither occur with the built-in completion styles like `basic` or `substring` nor with `orderless`. Please report the issue to `corfu-prescient`.

wyuenho:
This does [happen with orderless](https://github.com/oantolin/orderless/blob/master/orderless.el#L560), although, after some more testing, this seems to only happen to CAPF functions that return a category in the metadata, where the category contains the `flex` style. Which means, if one is using the built-in `eglot` version in Emacs 29, [is susceptible to this bug](https://github.com/emacs-mirror/emacs/blob/emacs-29/lisp/progmodes/eglot.el#L419). If one follows the [wiki's recommendation for LSP mode](https://github.com/minad/corfu/wiki#configuring-corfu-for-lsp-mode), one is also susceptible to this bug. In fact, if one overrides any of the categories for any of the CAPE functions to include a flex style, they'll run into this bug.

So, for safety, I'd recommend you merge this PR. There's no harm in doing so and saves a lot of people's time debugging and playing around with completion style overrides.

minad:
You are right, the `flex` style indeed doesn't use lazy highlighting in Emacs 29 and older. I pushed a fix.

wyuenho:
That also works. Let's see how well it works. Thanks.

PR #539, opened 2024-12-21
README: minor fix for example configuration

sctb:
Using :init to call ‘global-corfu-mode’ will fail if the package has not yet been loaded.

### DIFF
--- a/README.org
+++ b/README.org
@@ -130,7 +130,7 @@ Here is an example configuration:
   ;; Recommended: Enable Corfu globally.  This is recommended since Dabbrev can
   ;; be used globally (M-/).  See also the customization variable
   ;; `global-corfu-modes' to exclude certain modes.
-  :init
+  :config
   (global-corfu-mode))
 
 ;; A few more useful configurations...


minad:
Calling `global-corfu-mode` in `:init` should enforce loading Corfu.

sctb:
With the contents of my `init.el` as the following, I get an error:
```
(use-package corfu
  :load-path "corfu" ;; installed in my ~/.emacs.d
  :init (global-corfu-mode))
```
`⛔ Error (use-package): corfu/:init: Symbol’s function definition is void: global-corfu-mode`

Using `:config` works as expected. 

GNU Emacs 31.0.50 (build 1, aarch64-apple-darwin24.1.0, NS
 appkit-2575.20 Version 15.1.1 (Build 24B91)) of 2024-12-21

jdtsmith:
The something is wrong with your package installation, since `global-corfu-mode` is an autoload:

````elisp
;;; corfu-autoloads.el
(autoload 'global-corfu-mode "corfu" "\
Toggle Corfu mode in all buffers.
With prefix ARG, enable Global Corfu mode if ARG is positive;
otherwise, disable it....
```

Maybe try re-installing.

sctb:
Yes, I agree that autoload should make that function available in most use-cases. The PR was intended to suggest a change which works for strictly more cases (such as when `use-package` is invoked for loading and configuration only, not package installation). 

Additionally, the documentation for `use-package` states:
> Note that the ‘:init’ form is run unconditionally - even if the package happens to not exist on your system.  You must therefore remember to restrict ‘:init’ code to what would succeed either way; put the rest in the ‘:config’ section.
> ...
> In general, you should keep ‘:init’ forms as simple and quick as possible, and put as much as you can get away with into the ‘:config’ section.

In any case, my setup is working fine and I'm happy. Thanks for the bits!

minad:
> Yes, I agree that autoload should make that function available in most use-cases. The PR was intended to suggest a change which works for strictly more cases (such as when use-package is invoked for loading and configuration only, not package installation).

This is not correct. The `:config` stanza is only executed if the package is loaded, which means that `global-corfu-mode` will not be activated as intended.

SamB:
> This is not correct. The `:config` stanza is only executed if the package is loaded, which means that `global-corfu-mode` will not be activated as intended.
>
> EDIT: We could add a `(require 'corfu)` inside the `:init` section before `global-corfu-mode` to achieve the desired effect. But usually autoloads should be available. I wonder why there are not available in your setup.

Well, `use-package` only defers loading when you either ask it to do that or specify keywords that set up autoloads; see <https://www.gnu.org/software/emacs/manual/html_node/use-package/Deferring-loading.html>.

Notably, `use-package` **does not** do anything with `corfu-autoloads.el`. This is why `:init` does not have any autoloads in this case, but also means that `:config` is actually fine because `use-package` will load the package immediately.

I think it goes slightly differently if you use package.el instead of manual installation: in that case, package.el will load `corfu-autoloads.el` early, but `use-package` still doesn't care and will `(require 'corfu)` and eval `:config` without delay, except that `:init` was evaluated before that point, so if it triggered an autoload, there's nothing left for `require` to do.

PR #553, opened 2025-01-30
Use local image instead of links in README.org

manphiz:
* This makes it less of a round trip to get the images
* The patch was originally authored by Martin <debacle@debian.org>
  - https://salsa.debian.org/emacsen-team/emacs-corfu/-/commit/c569533e8f2e0bf51ed02ad7e83b88aeb38e759d


minad:
No, thanks. The package tarball for ELPA/MELPA is build from the repository and I don't want the images be part of it.

manphiz:
Understood.  Thanks for the explanation!

PR #567, opened 2025-04-05
Add corfu-history-duplicate

minad:
@jdtsmith Could you please check if this works as expected for you? Which value do you suggest? 0.5 as compromise?

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -2,6 +2,11 @@
 #+author: Daniel Mendler
 #+language: en
 
+* Development
+
+- ~corfu-history-duplicate~: Prioritize duplicate history elements when sorting by
+  history via ~corfu-history-mode~.
+
 * Version 2.0 (2025-03-28)
 
 - ~corfu-quick~: Ensure that popup does not move.
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -7,6 +7,9 @@
 - =corfu-history-duplicate= and =corfu-history-decay=: New customization options to
   adjust the rank of duplicate history elements, such that they appear earlier
   in the completion list.
+- =corfu-history-mode=: Ensure that the =corfu-history= is saved if =savehist-mode= is
+  enabled. Automatically add =corfu-history= to
+  =savehist-minibuffer-history-variables=.
 
 * Version 2.0 (2025-03-28)
 
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -4,8 +4,9 @@
 
 * Development
 
-- ~corfu-history-duplicate~: Prioritize duplicate history elements when sorting by
-  history via ~corfu-history-mode~.
+- =corfu-history-duplicate= and =corfu-history-decay=: New customization options to
+  adjust the rank of duplicate history elements, such that they appear earlier
+  in the completion list.
 
 * Version 2.0 (2025-03-28)
 
--- a/README.org
+++ b/README.org
@@ -127,11 +127,17 @@ Here is an example configuration:
   ;;        (shell-mode . corfu-mode)
   ;;        (eshell-mode . corfu-mode))
 
-  ;; Recommended: Enable Corfu globally.  This is recommended since Dabbrev can
-  ;; be used globally (M-/).  See also the customization variable
-  ;; `global-corfu-modes' to exclude certain modes.
   :init
-  (global-corfu-mode))
+
+  ;; Recommended: Enable Corfu globally.  Recommended since many modes provide
+  ;; Capfs and Dabbrev can be used globally (M-/).  See also the customization
+  ;; variable `global-corfu-modes' to exclude certain modes.
+  (global-corfu-mode)
+
+  ;; Enable optional extension modes:
+  ;; (corfu-history-mode)
+  ;; (corfu-popupinfo-mode)
+  )
 
 ;; A few more useful configurations...
 (use-package emacs
--- a/README.org
+++ b/README.org
@@ -525,7 +525,7 @@ following extensions come with the Corfu ELPA package:
 - [[https://github.com/minad/corfu/blob/main/extensions/corfu-echo.el][corfu-echo]]: =corfu-echo-mode= displays a brief candidate documentation in the
   echo area.
 - [[https://github.com/minad/corfu/blob/main/extensions/corfu-history.el][corfu-history]]: =corfu-history-mode= remembers selected candidates and sorts the
-  candidates by their history position.
+  candidates by their history position and frequency.
 - [[https://github.com/minad/corfu/blob/main/extensions/corfu-indexed.el][corfu-indexed]]: =corfu-indexed-mode= allows you to select indexed candidates with
   prefix arguments.
 - [[https://github.com/minad/corfu/blob/main/extensions/corfu-info.el][corfu-info]]: Actions to access the candidate location and documentation.
--- a/corfu.el
+++ b/corfu.el
@@ -886,7 +886,7 @@ See `completion-in-region' for the arguments BEG, END, TABLE, PRED."
   (setq beg (if (markerp beg) beg (copy-marker beg))
         end (if (and (markerp end) (marker-insertion-type end)) end (copy-marker end t))
         completion-in-region--data (list beg end table pred completion-extra-properties))
-  (completion-in-region-mode 1)
+  (completion-in-region-mode)
   (activate-change-group (setq corfu--change-group (prepare-change-group)))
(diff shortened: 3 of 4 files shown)


minad:
Now that I played with it for a while, and also with the corresponding Vertico implementation, I got second thoughts. The problem is that sometimes I want to repeat the most recent candidate, so the most recent candidate should always be ranked high. This is in conflict with the problem of recent candidates disturbing the list and repressing more frequent candidates. I think Prescient works like this - recent candidates have high priority and frequent candidates rather modify the end of the distribution. This is nice to have, but more of a fine tuning. In our case we also won't profit as much since the history duplicates won't provide the necessary precision such that we can profit from the fine tuning. Prescient stores frequency specially.

jdtsmith:
Just having a look. We could always add a special exception for the most recent single candidate, such that it can't be outranked.  I'd personally prefer to see more frequent selections ahead of "short" selections.  Not having played with it yet, I'd guess 1.0 is too high, maybe 0.2 or so would be my guess (with an enforced minimum of one index movement for repeaters).

> Duplicate elements in the history are prioritized such that they appear earlier in the completion list

This sounds like the weight applies equally for any duplicated once, twice, etc.  Maybe "repeated" instead of duplicate, and add "with more repeats increasing the priority".

Anyway, worth playing with it a bit to see if we like it.  No need to land immediately.

jdtsmith:
BTW I think you'd better guard against `dup=0` for non-zero `corfu-history-duplicate`; possibly just use the ceiling.

minad:
What do you think about https://github.com/minad/corfu/pull/567/commits/ea0b6ba4dc8ba3c09957e21eb9c517822fcb0ddd?

minad:
I think the new behavior where the most recent is never outranked is nice in theory. However I have my doubts that the behavior is actually noticeably better. It is the same with Prescient. I always had my doubts that frecency makes such a big difference. And then if this gets merged I wonder how much benefits Prescient will bring on top.

minad:
The corresponding Vertico branch is here in case you would also want to try that: https://github.com/minad/vertico/tree/frecency

minad:
> Yeah, it might be more of a feature for people who think they need frecency, but it basically adds one setting and very little code complexity.

Sure. But I don't like just adding something without evidence. People will still long for Prescient, because it yields even better frecency with a better formula for a better placebo effect.

> I thought I preferred a fractional "max migration" per repeat, but on reflecting further, this sorting elaboration only really matters if it "pulls up" something from off the initial page or two of candidates, since otherwise you'll probably reach for search instead.

One can always adjust the constant step size such that it corresponds to a fraction of the history size. I think a constant step size is a little easier to understand.

> A way to achieve that might be for all N candidates with history, to find those candidate's sequential history index position in the range i=[0..N-1], and move up by corfu-history-duplicate*f where f=(N-1-i)/(N-1) (or sqrt(f) if you think a gentler roll-off into the sunset is better).

Again, sounds good in theory and is easy to add. But diminishing returns?

> One additional problem that occurred to me with all this is that corfu-history is usually a mixed bag. 

Prescient also throws everything together. This is one more criticism I had against it. Note that in Vertico we have separate histories, only in Corfu everything is mixed. But yes, common names are desirable targets when completing after the dot.

minad:
Regarding losing rare candidates in the history, I think one can use the trick to only delete duplicates in the later half of the history list. This will give the best of both worlds. We get frecency analysis for recent enough candidates and we don't lose rare candidates. I find this important in shell completion with rarely executed commands.

~~~ elisp
(defun +history-delete-old-duplicates (var elt &rest _)
  (when-let ((old (nthcdr (/ history-length 2) (symbol-value var))))
    (setcdr old (delete elt (cdr old)))))
(advice-add #'add-to-history :before #+history-delete-old-duplicates)
~~~

Maybe this could be proposed as an Emacs feature too, e.g., a setting `history-delete-duplicates=old`.

jdtsmith:
Looking quite good to me in theory (haven't had time to play much).  If we call $\tau$ the `corfu-history-duplicate` shift length, maybe the weighted shift for history index $t$ is better expressed $w\equiv\tau e^{-t/(m\tau)}$.  I.e. it decays away in $m$ foldings of the chosen $\tau$ size (you have $m=20$ now).  Then you also know you need a history length $\gtrsim m\tau$ to fully harness the power of duplicate shifts.  

Would be great if `corfu-history` could be independent per `major-mode` (and `savehist` respected this).

jdtsmith:
I do like the idea of "retiring" old duplicates somehow, and proposing this as another setting for `history-delete-duplicates`.  PS: seems like `if/when-let*` is here to stay... sigh.

minad:
The problem with the `-t/(m*tau)` is that we cannot easily disable decay. Right now, setting `corfu-history-decay` to 0 will do. If we use `m` instead, then one has to set it to a large value. But I agree with you that it will be more naturally to express the decay like this.

> Would be great if corfu-history could be independent per major-mode (and savehist respected this).

I don't see a way except if we introduce tons of history variables. This is why I went with a single variable.

> I do like the idea of "retiring" old duplicates somehow, and proposing this as another setting for history-delete-duplicates. 

It is on my todo list for upstream, but it will be a little hard to argue without any benefits in Emacs itself.

> PS: seems like if/when-let* is here to stay... sigh.

One can use `(dolist (p '(if-let when-let)) (put p 'byte-obsolete-info nil))` in your config. I don't plan to change this soon. The point is that it leads to an unnecessary burden on many packages if commonly used language elements or macros are needlessly deprecated. It would have been better to alias when-let*=when-let.

minad:
Okay, I am going to merge this now. We can still adjust the decaying strategy if we learn more or if problems occur. Please let me know if you have further ideas!

jdtsmith:
> It might be more expected to decay consistently over a fixed number of elements, independent of history length.

Yes.  Assuming new history elements are added at a fixed rate per unit time during active work, this makes the most sense, even in the context of varying personal rates of history growth.

> we cannot easily disable decay

The variable could be settable to `nil` for no(/infinite) decay length.

> I don't see a way except if we introduce tons of history variables. This is why I went with a single variable.

Yeah.  Sounds like an idea for a general package that:

1. provides a way to specify history variables and the modes you'd like them to be unique in
2. sets them up buffer-local per mode in mode hooks
3. points the buffer-local version to the symbol plist (e.g. slot `per-mode-history-*mode*`) to .  I.e. make all `*mode*` buffers share the same list.[1]
4. hooks `savehist` to serialize all the plist slots as `some-history-var-*mode*` or similar, stashing the per-mode histories in `some-history-var`'s plist on read.  

[1] Not sure if shared bvars are possible.

jdtsmith:
>  It would have been better to alias when-let*=when-let

100% agree.  Changing established common symbols for purity reasons is never a good idea, IMO, unless they are a tripping hazard.

minad:
> The variable could be settable to nil for no(/infinite) decay length.

Yes, but this needs another conditional. Let's just keep it as is. To avoid decay just use a large value. Anyway I think we should find good defaults such that users don't have to change this.

> Yeah. Sounds like an idea for a general package that:

It is not clear to me if other variables would need a similar per-mode treatment. Given that Prescient also maintains only a single history, I don't see big reasons to support this. Either the candidates are different anyway or they are very common across programming languages. In the end the returns of separate histories will be small.

Anyway, I think the easiest way to implement this in Corfu would be to make corfu-history an alist of (mode . history-list). savehist-mode can also store this. The main downside is that we cannot use `add-to-history` anymore, and well also that the corfu-history variable would get multiple times bigger, maybe 10 times given that we can complete in many modes.

minad:
Have you tripped up much by the common history before? I mean the situation is not different now with the frecency support. I have not gotten any complaints about the shared history. Also in all the years in the case of Prescient the only one complaining was yours truly (https://github.com/radian-software/prescient.el/issues/85). So I think we could leave things as is and not overcomplicate. 🤷🏾

jdtsmith:
Hah, great minds think alike ;).  Much of my completion is aided by local mode-specific intelligence, e.g. eglot or iPython completions.  These "don't get confused" as they're based on known context, so joint history hasn't mattered much there.  But now that we're making old/common completions more mobile in the sorted list based on frequency, unwanted cross-mode promotions could become more of an issue.  For example, I complete `message` a lot in elisp, and so any python method named `message` will rocket to the front of the `obj.[Tab]` list.

> that the corfu-history variable would get multiple times bigger

I think the way to do it is to "opt-in" for unique history for separate (groups of) modes; otherwise use the shared global default.  But I'll see how it goes.

minad:
Yes, I see your point. Cross mode completions could become more confusing now with frequency taken into account. But then, I think Prescient is a pretty strong counter argument. Let's keep it for a while this and then discuss again. Please try it for a while in real life.

PR #586, opened 2025-08-11
Add escaping commands for corfu-separator

LuigiPiucco:
There are 2 new functions provided:

1. `corfu-insert-or-escape-separator` works exactly like Doom Emacs' +corfu/smart-sep-toggle-escape, meant to replace corfu-insert-separator when the user wants an insert/escape all-in-one solution. The code was refactored and looks very different, though.
2. `corfu-escape-separator` is provided as an addition to `corfu-insert-separator`, meant to be bound to another key. It never removes the separator, only the backslash. It may insert a separator if not there. This is for the user who wants to think of escaping separately from inserting, saving them the extra key to reinsert the separator if only option 1 was provided.

The functions' implementation is actually really simple, most of the diff is documentation.

Both are unbound by default. Maybe it would make sense use `corfu-insert-or-escape-separator` as the default bind, but that can come separately if need be.

By the way, if you think this functionality shouldn't be provided by the core of corfu, I don't mind if the PR gets closed. I jumped to a PR because this was a quick implementation and seems reasonable to add. I can shorten the documentation if it's too verbose, as well.

### DIFF
--- a/README.org
+++ b/README.org
@@ -393,6 +393,13 @@ Then, when a new orderless component is desired, use =M-SPC=
 and arbitrary orderless search terms and new separators can be entered
 thereafter.
 
+Alternatively, bind ~corfu-insert-or-escape-separator~ in place of
+~corfu-insert-separator~ to be able to escape and/or remove the separator
+indefinitely by repeating the bound key sequence. Or keep using
+~corfu-insert-separator~ and also bind ~corfu-escape-separator~, which only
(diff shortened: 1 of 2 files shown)


minad:
Thanks, but I think this doesn't fit here well. Corfu doesn't have any knowledge of an escaped separator character. The current behavior in Corfu is like this - press M-SPC to insert the separator and continue. If you want a literal space, just type SPC. Other separator characters can be configured similarly. Maybe document this in the wiki instead?

PR #621, opened 2026-04-10
Fix popup width for wide annotation glyphs

LuciusChen:
Candidates with wide annotation glyphs, such as color emoji shown on the
right-hand side, can get clipped by the popup on graphical displays.

The popup width is currently derived from character metrics via
`string-width`, which can underestimate the actual rendered width of some
glyphs. As a result, the popup frame can be sized too narrowly even though
the formatted candidate line looks wide enough logically.

This patch measures the rendered pixel width of the popup lines and uses
that as a lower bound for the popup width on graphical displays.

Since the final popup width can change after this adjustment, the horizontal
position is recomputed afterwards so the popup still stays within the
parent frame.

### DIFF
--- a/corfu.el
+++ b/corfu.el
@@ -763,6 +763,11 @@ FRAME is the existing frame."
                              (max 0 (+ corfu--index off 1 (- corfu-count))
                                   (min (- corfu--index off corr) corfu--scroll))))))
 
+(defun corfu--lines-pixel-width (lines)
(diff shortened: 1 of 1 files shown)


minad:
Thanks. Pixel perfect alignment has been discussed before - I don't want to call string-pixel-width since it is too inefficient. Therefore we have to hold on with this. On Emacs 31 the situation may improve and then we can reconsider.

PR #622, opened 2026-04-30
Support dynamic lookup of corfu-separator character

dadinn:
Make it possible to dynamically fetch separator character, such that buffer-local bindings can depend on other potentially buffer-local variables (e.g. taken from orderless-escapable-separator).


minad:
You can use a hook instead:

~~~ elisp
(add-hook 'corfu-mode-hook (lambda () (setq-local corfu-separator my-separator-var)))
~~~

dadinn:
Indeed, I've just read your comment [here](https://github.com/oantolin/orderless/pull/197#issuecomment-4351793377) , and realised that a hook would be perfectly sufficient, and this PR is not needed at all.

